PR #7, opened 2017-02-27
Help docs - adding content from the compass help section

rathisekaran:


kay-kim:
merged into local branch.  Did some minor edits and is in review with Thomas right now. Will merge and publish tomorrow.

PR #9, opened 2017-07-14
Updated image and field description of connect dialog: DOCSP-870

rob-mongodb:


source/includes/steps-starting-compass.yaml
```diff
@@ -15,7 +15,13 @@ content: |
        * - :guilabel:`Hostname`
 
          - The hostname of the machine where the ``mongod`` instance is
-           running.
+           running. 
+
+           .. tip::
+
+              Compass automatically populates other fields in this dialog
```
kay-kim: ?
If you enter a MongoDB connection string in this field, Compass automatically parses the string to populate relevant field in this dialog.

rob-mongodb: @kay-kim , it's not clear from your comment what you didn't like about the prior sentence.  I'm guessing that you wanted the dependent clause to come first? I just cut-and-pasted your sentence, but made "field" plural.  Staged at https://docs-mongodbcom-staging.corp.mongodb.com/compass/robertjustice/DOCSP-870/connect.html and ready for another look.

kay-kim: Sounds good. The reason I reversed is because this is the section where we're telling people to pop <x> into hostname field.  So, I thought it would be good to quickly get people to what values are acceptable:  The hostname, or here's a tip: enter the uri.

Also, I wanted people to know that we parse the URI to populate the other fields.

rob-mongodb:
@kay-kim , this is still waiting for your approval. Staged at https://docs-mongodbcom-staging.corp.mongodb.com/compass/robertjustice/DOCSP-870/connect.html.

PR #77, opened 2018-09-14
DOCS-11868, DOCSP-2586 - Connect page revisions and expand info for f…

jeff-allen-mongo:
…avorite connections

With these changes, we are tabbing out the installation procedure because the current installation steps are pretty overwhelming, so we want to try to only expose users to the info they want to see. If you have any opinions / other ideas for how this could be structured, let me know!

We are also cleaning up the Favorite Connections docs and making them a bit clearer / more verbose.

PS: This looks like a bigger change than it is, since a lot of the changes are coming from tabbing out the already existing table content.

Staged: https://docs-mongodbcom-staging.corp.mongodb.com/compass/jeffreyallen/DOCS-11868/connect.html


source/includes/fact-ssl-ssh-connection-warning.rst
```diff
@@ -0,0 +1,8 @@
+.. important::
+
+   Although |compass| supports connecting to a MongoDB instance through
+   an SSH tunnel via TLS/SSL, this configuration may lead to
+   unexpected behavior when connecting to a :manual:`secondary
+   </core/replica-set-secondary/>` member due to the possibility of the
+   :manual:`primary </core/replica-set-primary/>` changing during an
+   election.
```
skerschb: is the issue that if you are intentionally connecting to a secondary that secondary may become primary? This isn't completely clear to me.

source/includes/list-table-ssh-connection.rst
```diff
@@ -0,0 +1,178 @@
...
+              - ``id_rsa``
+
+              On Windows, the location of the identify files
+              depends on your choice of SSH client, such as PuTTY.
+
+          * - *SSH Passphrase*
+
+            - If your private key is encrypted, provide the
+              passphrase used to decrypt your private keys (stored
+              in the specified identity file). A passphrase
```
skerschb: keys == key

skerschb:
lgtm aside from one small nit. I think using a guide in the future would be great because then we could replace the existing connect via compass guide -- right now it's not as complete as this one. Any thoughts about using a guide template for this?

source/includes/fact-ssl-ssh-connection-warning.rst
jeff-allen-mongo: Yes, that's the issue as I'm understanding it. CC @imlucas on this in case there's something else here I'm missing.

source/includes/fact-ssl-ssh-connection-warning.rst
```diff
@@ -0,0 +1,9 @@
+.. important::
```
jeff-allen-mongo: Hey @imlucas, would you mind taking a look at this copy and ensuring this is accurate and properly describes the issues a user may run into when connecting using SSH and TLS/SSL together? I attempted to relay the issues from these tickets: 
https://jira.mongodb.org/browse/COMPASS-1930
https://jira.mongodb.org/browse/COMPASS-2796

source/includes/fact-ssl-ssh-connection-warning.rst
```diff
@@ -0,0 +1,9 @@
...
+
+   Although |compass| supports connecting to a MongoDB instance through
+   an SSH tunnel via TLS/SSL, this configuration may lead to
+   unexpected behavior when connecting to a :manual:`secondary
+   </core/replica-set-secondary/>` member. After an election, the
+   secondary member to which you are connected may become the
+   :manual:`primary </core/replica-set-primary/>`, which may lead to
+   |compass-short| not properly displaying your data.
```
skerschb: Does it not properly display the data or is it that the data might not be as expected? Would the same scenario play out if you connected to a primary directly (i.e., could become secondary)?

jeff-allen-mongo: @imlucas, can you speak to this?

imlucas: See the tickets linked above in https://github.com/mongodb/docs-compass/pull/77/files#r218224451

Yes, the same issues primary vs. secondary. Really, its just that the connection errors in a non-friendly way.

imlucas: Oh and yes on the data is weird tip: you could be looking at stale data, but the odd error case is much more likely

source/includes/fact-ssl-ssh-connection-warning.rst
imlucas: See below https://github.com/mongodb/docs-compass/pull/77/files#r220604110

PR #79, opened 2018-10-15
DOCSP-3494: Compass compatibility matrix mentioning 2.6

steveren:


jeff-allen-mongo:
Hey @steveren, there is one more reference to 2.6 on the FAQ page we should probably axe as part of this change.

steveren:
@jeff-allen-mongo done, thanks.

jeff-allen-mongo:
lgtm! I think we can go ahead and merge these in.

PR #80, opened 2018-10-19
DOCSP-3454 - Updating download URLs for new format

jeff-allen-mongo:


jdestefano-mongo:
Jeff, this should have used source-constants and two underscores. Here is what it should have looked like:

*conf.py*
Add `source-constants` to `extensions`. Then add this somewhere:

```
source_constants = {
  'downloads-page': '`downloads page <https://www.mongodb.com/downloads?jmp=docs#compass>`__'
}
```

Then update references in the docs to `{+downloads-page+}`

PR #86, opened 2018-10-26
(DOCSP-2594): Updates and refactors to Document Validation page

jeff-allen-mongo:
This is largely refactoring / reorganization. I also moved this page to under the Documents page in the ToC, since I think it makes more sense there.

Staged: https://docs-mongodbcom-staging.corp.mongodb.com/compass/jeffreyallen/DOCSP-2594/validation.html

### DIFF
--- a/source/validation.txt
+++ b/source/validation.txt
@@ -1,8 +1,8 @@
 .. _validation:
 
-===================
-Document Validation
-===================
+==========
+Validation
+==========
 
 .. default-domain:: mongodb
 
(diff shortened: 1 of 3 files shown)


source/validation.txt
```diff
@@ -22,60 +22,140 @@ Document Validation
    :ref:`MongoDB Compass Readonly Edition <compass-readonly>`.
    In Readonly Edition, validation rules may only be read.
 
-
 .. _validation-tab:
-.. _create-validation-rules:
 
 Validation Tab
 --------------
```
steveren: I'm curious why this page is titled "Document Validation" rather than "Schema Validation", as it is in the server manual. Either is OK, I think, but it might be better to stay consistent.

Also, it would be good to include a paragraph at the top of the page with a short explanation of what schema validation is, i.e. a mechanism for ensuring that all the documents in a collection conform to a certain shape or a certain range of values. The server manual page doesn't have this either, and I think it would be helpful.

PR #99, opened 2018-12-05
(DOCSP-3810): Compass does not support near / nearshpere

jeff-allen-mongo:
My original draft had an contextual sentence so this would read:

```
MongoDB Compass does not support the $near and $nearSphere geospatial query operators in any of the query bar fields. These operators do not provide any additional functionality in the :ref:`schema view <schema-tab>`.
```

But that didn't seem necessary. Just curious if you think it might help provide some context as to *why* we do not support these operators (at this time).

Staged: https://docs-mongodbcom-staging.corp.mongodb.com/compass/jeffreyallen/DOCSP-3810/query-bar.html

### DIFF
--- a/source/query-bar.txt
+++ b/source/query-bar.txt
@@ -12,8 +12,10 @@ Query Bar
    :depth: 1
    :class: singlecol
 
-From the various Collections tabs, you can use the query bar to specify
-MongoDB queries.  For details, see :ref:`querying`.
+From the :ref:`Documents <compass-documents>`, :ref:`Schema
+<schema-tab>`, and :ref:`Explain Plan <explain-plans>` tabs, you can
+use the query bar to specify MongoDB queries to return a subset of
+documents from the collection.
 
 .. figure:: /images/compass/query-bar-documents-view.png
    :figwidth: 720px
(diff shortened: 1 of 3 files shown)


source/query-bar.txt
```diff
@@ -268,24 +276,23 @@ Past Queries View
...
+:ref:`favorites <favorite-queries>` for the current collection.
 
 .. _recent-queries:
 
 Recent Queries
 ~~~~~~~~~~~~~~
 
-You can view up to 20 most recent queries for that collection. Click on
-a listed query to populate the query bar with that query.
+You can view up to the 20 most recent queries for that collection.
```
MongoCaleb: You can view up to 20 of the most recent queries...

source/query-bar.txt
```diff
@@ -56,6 +56,14 @@ options of the query.
 
      - The upper limit for the number of documents to return.
 
+.. note::
+
+   |compass| does not support the :manual:`$near
+   </reference/operator/query/near/>` and
+   :manual:`$nearSphere </reference/operator/query/nearSphere/>`
+   geospatial query operators in any of the query bar
+   fields.
```
MongoCaleb: I *do* think you should include that second sentence explaining why...or just append it with a comma: 
> ...any of the query bar fields, because they do not...

source/query-bar.txt
jeff-allen-mongo: Much better. Ty!

source/query-bar.txt
jeff-allen-mongo: Thanks for the input, reading it now I think it's good to include.

PR #104, opened 2019-01-31
(DOCSP-4315): Landing page cleanup

jeff-allen-mongo:
Cleaned up the copy on the landing page and separated the Contact section to a new page.

I also added an image to the landing page, so users have an idea of what Compass actually looks like. Do we like this? Yay or neigh 🐴 ? I wasn't sure if it looked best before or after the Overview section.

Staged view: https://docs-mongodbcom-staging.corp.mongodb.com/compass/jeffreyallen/DOCSP-4315/index.html

New Feedback page: https://docs-mongodbcom-staging.corp.mongodb.com/compass/jeffreyallen/DOCSP-4315/submit-feedback.html

### DIFF
--- a/source/index.txt
+++ b/source/index.txt
@@ -9,43 +9,41 @@ MongoDB Compass
 .. contents:: On this page
    :local:
    :backlinks: none
-   :depth: 1
+   :depth: 2
    :class: singlecol
 
-Overview
---------
+What is |compass-short|?
+------------------------
 
-.. admonition:: |compass| is available in the following editions
+|compass| is the :abbr:`GUI (Graphical User Interface)` for MongoDB.
+|compass-short| allows you to analyze and understand the contents of
+your data without formal knowledge of MongoDB
+:manual:`query syntax </tutorial/query-documents>`. In addition to
+exploring your data in a visual environment, you can also use
+|compass-short| to :ref:`optimize query performance <explain-plans>`,
+:ref:`manage indexes <compass-indexes>`, and implement
+:ref:`document validation <validation>`.
 
-   - Compass
+Compass Editions
+----------------
 
-   - Compass Community Edition
+- Compass
 
-   - Compass Readonly Edition (*New in version 1.12.0*)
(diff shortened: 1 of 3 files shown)


gilberto-mdb:
Hi Jeff,

I think we can remove the current screenshot. It shows a lot of information at once and this page may be / is the first entry point for a lot of potential Compass users, which could confuse or intimidate them / drive them away. I am, however, working with a visual designer to add a visual to this page, much like the Stitch docs landing page. There is a good chance we will reuse the screenshot you added somewhere else in the body of this documentation (TBD), but likely not in the overview, so it can be removed.

We can also remove the second paragraph in the "Overview" section since it refers to the screenshot / is not needed (in retrospect). The tidying up of the first paragraph looks nice.

Thank you for your work on this!

jeff-allen-mongo:
Hey @gilberto-mdb, I updated with your suggestions. I also removed the _Overview_ heading. This is a practice we're adopting across the CET docs since it's a bit more minimalist, and the opening section of a page is always understood to be an Overview.

gilberto-mdb:
Got it. Thanks for the context. It looks cleaner, but also muted because of the large ToC above it and the "Compass Editions" heading right below it. It's pretty important information that is worth calling out. If we are removing "Overview" generally across Docs, what about something like "What is Compass?" as the heading here? I think that would avoid "Overview" but still do a good job of calling out the information.

gilberto-mdb:
@jeff-allen-mongo 

On the new feedback page, can we put the last three sentences into a bulleted list, which each sentence being its own bullet please?

jeff-allen-mongo:
@gilberto-mdb, thanks for the suggestions. Changes have been made and this is ready for another look.

gilberto-mdb:
Looks good to me! @jeff-allen-mongo 

Adding what we discussed in person:

- Jeff will merge changes in so that Steve and Tony can work on the remaining tickets
- Jeff will save original versions of the landing page so that we can show and tell changes to the page in the future.

PR #105, opened 2019-02-05
DOCSP-4338: [Compass] Call out differences in Compass editions using a feature table

steveren:
Staged: https://docs-mongodbcom-staging.corp.mongodb.com/compass/steverenaker/DOCSP-4338/index.html

### DIFF
--- a/source/index.txt
+++ b/source/index.txt
@@ -12,23 +12,10 @@ MongoDB Compass
    :depth: 1
    :class: singlecol
 
+
+
 Overview
 --------
-
-.. admonition:: |compass| is available in the following editions
-
-   - Compass
-
-   - Compass Community Edition
-
-   - Compass Readonly Edition (*New in version 1.12.0*)
-
-   - Compass Isolated Edition (*New in version 1.14.0*)
-
-   Each |compass| edition is available in both a
-   *General Availability* and a *Beta* version. To download any
-   available version, go to the `MongoDB Download
-   Center <https://www.mongodb.com/downloads?jmp=docs#compass>`_.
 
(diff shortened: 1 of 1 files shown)


gilberto-mdb:
Hi @steveren 

Here's my feedback:

- The feature table is a lot longer (vertically) than I had imagined it would be. There seems to be some overlap in some of the rows as well. Can you take another pass and collapse (join) some of the rows to make the table shorter please? For example, row 1 and row 3 can probably be combined into one row that says "View databases, collections, and documents" (row 1 and 3 are features shared across the exact same editions). We could also probably collapse the three rows re: Authentication into one, I imagine, and say something like: "Authentication (Kerberos, LDAP, x509)". The row "Intercom" can be removed as it's more of a support feature rather than a Compass-specific feature. These aren't all the collapsible rows, but these stood out / came to mind. 

- The ticket didn't originally indicate to add a title over the feature chart, but I am very glad you did. I think it's very important to call out this chart to increase scannability for users and help them make decisions about which edition is right for them. I think we can be a bit more descriptive with the title though. Can we please change that title to: "What are the capabilities of each Compass Edition?" Also, please modify the second sentence in the "Available Compass Editions" section as needed (currently it references a specific "feature chart").

No action required on this, but just want to let you know that I'm wondering whether the information is the "Available Compass Editions" section is better consumed via table. Will keep thinking on it...

steveren:
@gilberto-mdb thanks for the feedback. I made some tweaks to the feature table to consolidate some of the rows and renamed the comparison table section.

steveren:
This PR goes to the wrong branch; closing.

PR #113, opened 2019-02-14
S&T: adds links to table topics, minor rewording

schmalliso:
Some wordsmithing, link additions, and reordered one item from the table.

I still have no idea what "This version disables all network connections except the connection to the MongoDB instance." means, so I think that needs clarification.


schmalliso:
S&T doc: https://docs.google.com/document/d/1gn5xhg4SVUWMKNDYO_5j8cm_1g3XKLk7xMMmuLK2KkI/edit

PR #133, opened 2019-03-29
(DOCSP-4644): Connection Troubleshooting

jeff-allen-mongo:
A note - A few of Brian's Troubleshooting articles largely covered the same issue or did not really make sense in a "Troubleshooting" context, so not all of his articles / topics are covered in this page.

Something I'd like to consider is adding a separate page covering the content from his [Connecting to a Primary or Secondary](https://intercom.help/mongodb-compass/connect-with-compass/connecting-to-a-replica-set-primary-or-secondary) article, possibly also under Troubleshooting.

Staged: https://docs-mongodbcom-staging.corp.mongodb.com/compass/jeffreyallen/DOCSP-4644/troubleshooting.html


config/sphinx_local.yaml
```diff
@@ -9,6 +9,8 @@ theme:
   google_analytics: 'UA-7301842-8'
   nav_excluded:
     - /
+    - /troubleshooting
```
gilberto-mdb: Do you think we need this landing page? I've seen them throughout docs before, but not sure if they are standard or anything like that.

source/troubleshooting/connection-errors.txt
```diff
@@ -0,0 +1,311 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+MongoDB Not Running on the Provided Host and Port
+-------------------------------------------------
+
+This error indicates that when |compass-short| attempted to connect to
```
gilberto-mdb: How about?

> Your connection failed when Compass attempted to connect to your database on the specified host and port.

Just trying to be a bit more scannable / shorter.

source/troubleshooting/connection-errors.txt
```diff
@@ -0,0 +1,311 @@
...
+~~~~~~~~~~~~~~~~~~~~~
+
+When you are in this state:
+
+- The :guilabel:`Create Collection` and :guilabel:`Insert Document`
+  buttons are disabled.
+
+- You cannot modify any of your documents.
+
+Likely Cause
```
gilberto-mdb: I know this is silly, but what if people don't know they downloaded the Readonly version of Compass? Should we say anything about that and possible link back to the landing page so that they can learn a bit more about the different editions of Compass?

source/troubleshooting/connection-errors.txt
```diff
@@ -0,0 +1,311 @@
...
+~~~~~~~~~~~~
+
+You are likely connected to a secondary member of your cluster, which
+does not support write operations.
+
+Solutions
```
gilberto-mdb: Just adding a note here since I added the previous comment re: Readonly Compass.

source/troubleshooting/connection-errors.txt
```diff
@@ -0,0 +1,311 @@
...
+
+- You have not :atlas:`configured any MongoDB users
+  </security-add-mongodb-users/>` for your |service| cluster.
+
+- You did not specify the correct :guilabel:`Authentication Database`.
+  This setting corresponds to the :manual:`authSource
+  </reference/connection-string/#urioption.authSource>` connection
+  string option.
+
+  .. note::
```
gilberto-mdb: Not sure a note is needed for this. I think we can do 1 of 2 things:

- Move bullet point 4 (and its accompanying note) to be bullet point 5 (so that we don't chop bullet point 5 away from the list, as it currently looks)

- Remove the Note and keep everything else as-is

gilberto-mdb:
On your note here @jeff-allen-mongo:

>Something I'd like to consider is adding a separate page covering the content from his Connecting to a Primary or Secondary article, possibly also under Troubleshooting.

So what do you imagine the section on the left would look like?

Troubleshooting
- Compass Connection Errors
- Connecting to a Primary or Secondary

?

config/sphinx_local.yaml
jeff-allen-mongo: We don't really have a standardized practice here. I was thinking it may be useful if we eventually add other types of Troubleshooting pages (in addition to Connection Troubleshooting). However I agree that right now it looks a bit weird with just one page. I can update to make it a single top-level page. If we move this to a single page, it may make sense to move this to a subpage under Connect to MongoDB. What are your thoughts?

jeff-allen-mongo:
@gilberto-mdb, yes that is along the lines of what I was thinking.

config/sphinx_local.yaml
gilberto-mdb: Let me make sure I understand:

It's required that all sections in the left hand side of our docs have an associated landing page. It's not possible to just click "Troubleshooting" and have that expand to show the subpages. Is that right? 

This makes sense to me:

> I was thinking the top-level landing page may be useful if we eventually add other types of Troubleshooting pages (in addition to Connection Troubleshooting). However I agree that right now it looks a bit weird with just one page. I can update to make it a single top-level page.

This part doesn't make sense to me:

> If we move this to a single page, it may make sense to move this to a subpage under Connect to MongoDB.

I think we'll need an explicit section for troubleshooting. Shall we discuss IRL?

source/troubleshooting/connection-errors.txt
gilberto-mdb: 🎉

source/troubleshooting/connection-errors.txt
gilberto-mdb: Looks great

source/troubleshooting/connection-errors.txt
gilberto-mdb: Yep I'm fine with that. As I look at it again, I think it just doesn't look apparent that the note is associated with bullet point 5 (arguably). I think what makes it potentially confusing is that we use the word "this" in the note, so the context may be lost if someone doesn't know that the note corresponds to bullet point 5 (i.e., "What is 'this'?")

gilberto-mdb:
On your note re: 

> @gilberto-mdb, yes that is along the lines of what I was thinking.

That makes sense to me and I'd be fine with that. @jeff-allen-mongo

config/sphinx_local.yaml
jeff-allen-mongo: Discussed IRL. Leaving as-is for now since we need a landing page for top-level pages. Going to remove the previous / next page arrows. This will look better once we add a second page.

source/troubleshooting/connection-errors.txt
jeff-allen-mongo: Agreed. Will update to keep the context.

gilberto-mdb:
@jeff-allen-mongo LGTM!

sampoonachot:
1. More than half of the external links don't work? I see a 404 error

2. I don't think we need an external link icon next to "http://portquiz.net:27017/." because it has 'http' in the text. 

3. Should the SEE ALSO message live in a green (or other color) banner? similar to NOTE?

jeff-allen-mongo:
Thanks @sampoonachot!

1. Oops that's my mistake. Had bad formatting for the Atlas URL's but should be good now!

2. Unfortunately we don't have a choice except to show the ext link icon (it's automatically generated by our toolchain).

3. I just tried re-staging with the content in a colored banner and found it looked weird in the [Allow to Connect to Secondary](https://docs-mongodbcom-staging.corp.mongodb.com/compass/jeffreyallen/DOCSP-4644/troubleshooting/connection-errors.html#allow-compass-to-connect-to-the-secondary) section with two colored boxes right after each other.

PR #144, opened 2019-06-12
(DOCSP-5705): Refactor Import / Export page to use step files (in progress)

tahiyachowdhury:
Jira: https://jira.mongodb.org/browse/DOCSP-5705?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel

Staging: https://docs-mongodbcom-staging.corp.mongodb.com/compass/tahiyachowdhury/DOCSP-5705/import-export.html#procedure

### DIFF
--- a/source/includes/steps-export.yaml
+++ b/source/includes/steps-export.yaml
@@ -1,0 +1,68 @@
+title: Navigate to the :doc:`collection </collections>` you wish to export data into.
+level: 4
+ref: select-collection
+content: |
+  You can either select the collection from the
+  :doc:`Collections </collections>` tab or click the collection in the
+  left-hand pane.
+
+---
+title: Click :guilabel:`Collection` in the top-level menu and select :guilabel:`Export Data`
+level: 4
+ref: export-data
+content: |
+  :guilabel:`Export Collection`:
+
+  .. raw:: html
+
+     <br>
+
+  .. figure:: /images/compass/export-data-option-select.png
+     :alt: Export data select menu
+
+  .. raw:: html
+
+     <br>
+
+  |compass-short| displays the following dialog:
+
+  .. raw:: html
+
+     <br>
+
+  .. figure:: /images/compass/export-data-dialog.png
+     :alt: Export data dialog
+
+  The top section of the export dialog displays the query
+  entered in the query bar prior to export, if applicable. You
+  can use this query to filter which collection documents are
+  exported. If no query was specified, this section displays
+  ``undefined``.
+
+---
+title: |
+  Use the :guilabel:`Export Full Collection` toggle to indicate 
+  whether to export only the documents matched by the query or the
+  full collection.
+
+level: 4
+ref: export-collection
+content: |
+  .. note::
+     Having an ``undefined`` query filter and toggling the
+     option to export the full collection both result in
+     |compass-short| exporting the full collection.
+
+---
+title: Choose the appropriate file type and the location of the export file.
+level: 4
+ref: click-file-location
+content: ""
+---
+title: Click :guilabel:`Export`. 
+level: 4
+ref: click-export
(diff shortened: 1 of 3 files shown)


jeff-allen-mongo:
Hey @tahiyachowdhury, just wondering did you update staging with the latest version?

1. `make html`

2. `make stage`

It looks like what's currently in staging has some formatting issues:
Screen Shot 2019-06-12 at 2 20 43 PM

source/includes/steps-export.yaml
```diff
@@ -0,0 +1,70 @@
+title: Navigate to the :doc:`collection </collections>` you wish to export data into.
```
atsansone:
```suggestion
title: Navigate to the :doc:`collection </collections>` you wish to export data into.
stepnum: 1
```

source/includes/steps-export.yaml
```diff
@@ -0,0 +1,70 @@
...
+  ``undefined``.
+
+---
+title: Use the :guilabel:`Export Full Collection` toggle to indicate 
+  whether to export only the documents matched by the query or the
+  full collection.
+level: 4
+ref: export-full-collection
+content: |
+  .. note::
```
jeff-allen-mongo: So I'd say that we maybe no longer need this in a note. Since it now lives below a clearly defined step and is providing context for the step, it might be able to live in plain copy. I think either way works, so I'll leave it to you to decide.

source/includes/steps-export.yaml
```diff
@@ -0,0 +1,70 @@
...
+     Having an ``undefined`` query filter and toggling the
+     option to export the full collection both result in
+     |compass-short| exporting the full collection.
+
+---
+title: Choose the appropriate file type and the location of the export file.
```
jeff-allen-mongo: Thought: Maybe we split this into two steps since it's separate in the UI. We could maybe have one step for choosing the file type, and provide a context block that users can choose either csv or json. We could then have a second step for the export location.

source/includes/steps-export.yaml
```diff
@@ -0,0 +1,77 @@
...
+  ``undefined``.
+
+---
+title: Use the :guilabel:`Export Full Collection` toggle to indicate 
+  whether to export only the documents matched by the query or the
+  full collection.
+level: 4
+ref: export-full-collection
+content: |
+    Having an ``undefined`` query filter and toggling the
```
jeff-allen-mongo: Two things here - so I actually think it's easier to read in note form. It looks a little scrunched as is and tough to scan. Sorry for going back and forth on this.

Also, I think we can phrase this to be more active voice. Consider something like:

```
|compass-short| exports the full collection both when the query filter is ``undefined`` and when you enable the
:guilabel:`Export Full Collection` option.
```

source/includes/steps-export.yaml
```diff
@@ -0,0 +1,77 @@
...
+  .. raw:: html
+
+     <br>
+
+  .. figure:: /images/compass/export-data-dialog.png
+     :alt: Export data dialog
+
+  The top section of the export dialog displays the query
+  entered in the query bar prior to export, if applicable. You
+  can use this query to filter which collection documents are
```
jeff-allen-mongo: [nit] we can remove "You can" from this sentence to promote a more active tone.

source/includes/steps-export.yaml
jeff-allen-mongo: actually, maybe say 

```
|compass-short| exports the full collection when the query filter is ``undefined``, or when you enable the :guilabel:`Export Full Collection` option.
```

my initial suggestion makes it sound like you need both to be true.

tahiyachowdhury: Good catch! Definitely easier to read now.

PR #153, opened 2019-07-24
(DOCSP-5918): Support for wildcard indexes

jeff-allen-mongo:
Compass supports creating wildcard indexes. There is a bit of refactoring done as part of this change. Now, instead of linking the index options and linking out to more information, we include a brief one-line description of each option. We still link out to the manual for more info.

JIRA: [DOCSP-5918](https://jira.mongodb.org/browse/DOCSP-5918)

Staged: [Index Options](https://docs-mongodbcom-staging.corp.mongodb.com/compass/jeffreyallen/DOCSP-5918/indexes.html#optional-specify-the-index-options)

### DIFF
--- a/source/includes/steps-create-index.yaml
+++ b/source/includes/steps-create-index.yaml
@@ -17,7 +17,7 @@ title: Add fields to index.
 level: 4
 ref: index-fields
 content: |
-   To specify a key for the index, select the field and the index type
+   To specify a key for the index, select the field and the index type.
    To index additional fields, click :guilabel:`Add Another Field`.
 ---
 title: Optional. Specify the index options.
@@ -37,8 +37,8 @@ content: |
 
       * - Build index in the background
 
-        - If checked, the MongoDB deployment remains available during
-          the index build operation.
+        - If checked, ensure that the MongoDB deployment remains
+          available during the index build operation.
 
         - :manual:`Background Construction </core/index-creation/index.html#background-construction>`
 
@@ -72,8 +72,8 @@ content: |
 
       * - Wildcard projection (*New in MongoDB 4.2*)
         
-        - If checked, the index supports unknown or arbitrary fields
-          which match the specified projection.
+        - If checked, support unknown or arbitrary fields
+          which match the specified projection in the index.
 
         - :manual-next:`Wildcard Indexes </core/index-wildcard/>`
 ---
--- a/source/includes/steps-create-index.yaml
+++ b/source/includes/steps-create-index.yaml
@@ -27,18 +27,55 @@ content: |
 
    |compass-short| supports the following index options:
 
-   * `Build the index in the background <https://docs.mongodb.com/master/core/index-creation/index.html#background-construction>`_
+   .. list-table::
+      :header-rows: 1
+      :widths: 40 60 20
 
-   * :manual:`Create a unique index </core/index-unique>`
+      * - Option
+        - Description
+        - More Information
 
-   * :manual:`Create a TTL Index </core/index-ttl>`
+      * - Build index in the background
 
-   * :manual:`Create a partial index </core/index-partial/>`
+        - If checked, the MongoDB deployment remains available during
+          the index build operation.
 
-   * `Use custom collation <https://docs.mongodb.com/manual/indexes/index.html#indexes-and-collation>`_
+        - :manual:`Background Construction </core/index-creation/index.html#background-construction>`
 
-     For information on the Custom Collation settings, see :manual:`Collation Document </reference/collation/#collation-document>`
-     in the MongoDB manual.
+      * - Create unique index
+
+        - If checked, ensure that the indexed fields do not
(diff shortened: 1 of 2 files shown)


source/includes/steps-create-index.yaml
```diff
@@ -27,18 +27,55 @@ content: |
```
melissamahoney-mongodb: I don't know how to comment on the rest of this file? But Step 3 is missing a period between the two sentences.

source/includes/steps-create-index.yaml
```diff
@@ -27,18 +27,55 @@ content: |
...
+        - :manual:`Partial Indexes </core/index-partial/>`
+
+      * - Use custom collation
+
+        - If checked, create a custom collation for the index
+          using the options provided in |compass-short|.
+
+        - :manual:`Collation Document </reference/collation/#collation-document>`
+
+      * - Wildcard projection (*New in MongoDB 4.2*)
```
melissamahoney-mongodb: See above about parallelism. Possibly: "Use wildcard projection".

Nit: I wonder if we could put a line break between "Wildcard projection" and "(New in MongoDB 4.2.)"?

source/includes/steps-create-index.yaml
```diff
@@ -27,18 +27,55 @@ content: |
...
+        - :manual:`Unique Indexes </core/index-unique>`
+
+      * - Create :abbr:`TTL (Time to Live)`
+
+        - If checked, automatically delete documents after a
+          specified number of seconds since the indexed field value.
+
+        - :manual:`TTL Indexes </core/index-ttl>`
+
+      * - Partial filter expression
```
melissamahoney-mongodb: Are these options in the UI or did you name them? If possible, it would more consistent if these were all syntactically parallel. For example: "Use partial filter expression".

source/includes/steps-create-index.yaml
```diff
@@ -27,18 +27,55 @@ content: |
...
+      * - Use custom collation
+
+        - If checked, create a custom collation for the index
+          using the options provided in |compass-short|.
+
+        - :manual:`Collation Document </reference/collation/#collation-document>`
+
+      * - Wildcard projection (*New in MongoDB 4.2*)
+        
+        - If checked, the index supports unknown or arbitrary fields
```
melissamahoney-mongodb: Parallelism with the other descriptions, if possible. Maybe: "If checked, support unknown or arbitrary fields which match the specified projection in the index". ?

source/includes/steps-create-index.yaml
```diff
@@ -27,18 +27,55 @@ content: |
...
-   * :manual:`Create a unique index </core/index-unique>`
+      * - Option
+        - Description
+        - More Information
 
-   * :manual:`Create a TTL Index </core/index-ttl>`
+      * - Build index in the background
 
-   * :manual:`Create a partial index </core/index-partial/>`
+        - If checked, the MongoDB deployment remains available during
```
melissamahoney-mongodb: This description isn't quite parallel with the others.

source/includes/steps-create-index.yaml
jeff-allen-mongo: Fixed!

source/includes/steps-create-index.yaml
jeff-allen-mongo: These are all how the options are shown in the UI. I debated whether it would be better to just go with what is in the UI vs name them ourselves. Ultimately I went with what is in the UI, but I can talk with product and see if they are opening to making these options more parallel in the UI.

source/includes/steps-create-index.yaml
jeff-allen-mongo: Agreed. Tweaked a bit to try to make it a bit more parallel.

source/includes/steps-create-index.yaml
jeff-allen-mongo: 1. ibid my other comment about these names matching what is shown in the UI, but will discuss with product.

2. 
Screen Shot 2019-07-24 at 4 37 36 PM unfortunately it looks like this is the only way we can get it to show up with a line break, which looks a bit weird imo. I think it's not ideal but having it on the same line looks a bit better. What do you think?

source/includes/steps-create-index.yaml
jeff-allen-mongo: Agreed. Updated.

source/includes/steps-create-index.yaml
melissamahoney-mongodb: I figured. And I think its right to go with whats in the UI, prevents confusion. But see if can you can maybe bend product's ear ;)

source/includes/steps-create-index.yaml
melissamahoney-mongodb: Ah, ok. I agree to keep it on the same line. Thanks for trying!

PR #206, opened 2019-12-13
(DOCSP-7838): Making querying documents more task-centric (ToC Refactor)

jeff-allen-mongo:
This is a refactor of some parts of the Compass docs to make them more task-centric rather than feature-centric. The goal of this refactor is for users who come to the docs with a task they want to perform, it is easy to find instructions on how to do that task.

For context, before this change, querying data was nested deeply under `Collections and Views` > `Query Bar` > `Filter Returned Documents`. This page is really difficult to get to if you're just using the ToC.

Now, it should be more intuitive to get to this page by going from `Interact with your data` to `Query your data`.

For this review, I'm looking for feedback on the content _structure_, not so much the content itself. I'd like to know if you think this is a good way to organize the content, or if there are any other improvements we can make.

Current live Compass docs: https://docs.mongodb.com/compass/current/
Staged content refactor: https://docs-mongodbcom-staging.corp.mongodb.com/compass/docsworker/DOCSP-7838-new/index.html


source/databases.txt
```diff
@@ -30,14 +30,6 @@ database collections by clicking the desired database in the left-hand navigatio
 You can also :ref:`create <create-database>` or
```
gilberto-mdb: Is the Databases page new?

source/collections.txt
```diff
@@ -75,14 +75,6 @@ a unique icon:
```
gilberto-mdb: I really like the "Databases" page at the top of this section. I do think it's quite natural to start talking about Collections and Views after you talk about databases (assuming someone is following the pages in linear top-down order), but I think the way the "Collections and Views" subsection is currently structured is a bit strange. Specifically:

- There's a "Collections and Views" _section_, but when you expand it it's only one page, which feels unnecessary. 

- The "Collections and Views" _page_ is titled "Collections and Views" but the vast majority of the content on the page is about collections, with one small section on "Views", and I think that section is redundant with the section on the actual "Views" page.

Some _possible_ changes to consider:

- Remove the "Views" section in the "Collections and Views" page and migrate it to the "Views" page, if necessary. 

- Remove the "Collections and Views" sections and just have single, separate "Collections" and "Views" pages, much like the "Databases" page. These two pages would go right underneath the "Databases" page in the following order: "Databases", "Collections", "Views".

- If we're really tied to the idea of the ">" expandable subsection, and want to double down on task-oriented documentation, it's possible we could create a  "> Collections" expandable subsection, and within that subsection, have individual pages that correspond to the sections in the "Collections and Views" page. So for example, you'd have ">Collections" as the expandable menu and then you'd have "Create a Collection", "Drop a Collection", etc. (like the "> Manage Documents" section. The pages in ">Collections" would be quite short though, but perhaps that's not that big of a deal.

source/manage-data.txt
```diff
@@ -0,0 +1,20 @@
+=======================
+Interact with Your Data
```
gilberto-mdb: Perhaps not as part of this ticket, but in the future, we should consider:

- Finding a different, consolidated home in the Compass docs for the "Databases" page, the "Collections and Views" pages, and the "Manage Documents" pages (just a thought).

- Promoting all of the pages under the "Query Your Data" and "View Past Queries" expandable subsections to be "top level" pages, meaning that the final version of this ToC section would be "Interact with Your Data" followed by "Set Which Fields Are Returned" down through "Set Validation Rules for Your Schema". Under this organization, there'd be no need for expandable subsections with "Interact with Your Data".

Again, not as part of this ticket, but perhaps something to consider in the future so that we can (1) continue to improve the organization of the docs whenever possible (based on what users are doing with Compass) and (2) highlight a "task-oriented" experience in the docs as much as possible. The Compass docs have so much potential to be _the_ model of what task-oriented doc'n looks like (these docs are doing an amazing job right now) and I'd like to evangelize them as much as possible.

jeff-allen-mongo:
Thanks @gilberto-mdb, ready for another look!

source/databases.txt
gilberto-mdb: Cool - just curious

PR #212, opened 2020-01-23
(DOCSP-8369) Explain what indexes are on the Manage Indexes page

zach-carr:
[Ticket](https://jira.mongodb.org/browse/DOCSP-8369)
[Staged](https://docs-mongodbcom-staging.corp.mongodb.com/compass/zach.carr/DOCSP-8369/indexes.html)

Added an untitled overview section that provides some context for a user who doesn't know what indexes are or how to use them. It's meant to frame the page or get them started reading the server docs.

Added an admonition calling out potential tradeoffs.

### DIFF
--- a/source/indexes.txt
+++ b/source/indexes.txt
@@ -12,6 +12,8 @@ Manage Indexes
    :depth: 1
    :class: singlecol
 
+To learn about Indexes, see :manual:`Indexes </indexes>`.
+
 .. _indexes-tab:
 
 Indexes Tab
--- a/source/indexes.txt
+++ b/source/indexes.txt
@@ -21,10 +21,10 @@ To improve query performance, build indexes on fields that appear often
 in queries and for all operations that 
 :ref:`sort by a field <query-bar-sort>`.
 
-- Queries on an indexed field use the index to limit the number of 
+- Queries on an indexed field can use the index to limit the number of 
   documents that must be scanned to find matching documents.
 
-- Sort operations on an indexed field return documents pre-sorted 
+- Sort operations on an indexed field can return documents pre-sorted 
   by the index.
 
 To learn more about indexes, see :manual:`Indexes </indexes>`.
--- a/source/indexes.txt
+++ b/source/indexes.txt
@@ -12,16 +12,20 @@ Manage Indexes
    :depth: 1
    :class: singlecol
 
-Indexes are special data structures that support the efficient 
-execution of queries in MongoDB. Indexes store a portion of a 
-collection's data in an easy-to-traverse form, thereby supporting 
-efficient equality matches and sort operations.
+Indexes are special data structures that improve query performance.
+Indexes store a portion of a collection's data in an easy-to-traverse 
+form. The index stores the value of a specific field or set of fields, 
+ordered by the value of the field.
 
 To improve query performance, build indexes on fields that appear often 
 in queries and for all operations that 
-:ref:`sort by a field <query-bar-sort>`. Queries on an indexed field 
-use the index to limit the number of documents that must be scanned or 
-to return a set of documents pre-sorted by that index.
+:ref:`sort by a field <query-bar-sort>`.
+
+- Queries on an indexed field use the index to limit the number of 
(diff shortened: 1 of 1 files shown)


source/indexes.txt
```diff
@@ -12,6 +12,26 @@ Manage Indexes
...
    :class: singlecol
 
+Indexes are special data structures that store a small portion of a 
+collection’s data set in an easy-to-traverse form. Indexes support the 
+efficient execution of queries in MongoDB.
+
+To improve query performance, build indexes on fields that appear often 
+in queries and for all operations that sort by a field. Queries on the 
+indexed field use the index to limit the number of documents that must 
+be scanned.
```
jeff-allen-mongo: So in the first sentence we talk about queries and sort operations. The next sentence explains why you should build an index for queries. Maybe should add a sentence or tweak the first to explain why you want one for sort operations?

source/indexes.txt
```diff
@@ -12,6 +12,28 @@ Manage Indexes
    :depth: 1
    :class: singlecol
 
+Indexes are special data structures that support the efficient
```
jeff-allen-mongo: "support the efficient execution of queries in MongoDB" sounds like too much passive voice. What about:

"Indexes are special data structures that improve query (execution|performance)."

source/indexes.txt
```diff
@@ -12,6 +12,28 @@ Manage Indexes
...
    :class: singlecol
 
+Indexes are special data structures that support the efficient 
+execution of queries in MongoDB. Indexes store a portion of a 
+collection's data in an easy-to-traverse form, thereby supporting 
+efficient equality matches and sort operations.
+
+To improve query performance, build indexes on fields that appear often 
+in queries and for all operations that 
+:ref:`sort by a field <query-bar-sort>`. Queries on an indexed field
```
jeff-allen-mongo: Suggest making this two bullets below "sort by a field". Something like:

```
- Queries on an indexed field can use the index to limit the number of documents
  that must be scanned to return the matching documents.

- Sort operations on an indexed field can use the pre-sorted results provided by the index.

File: source/indexes.txt [hunk 4]
```diff
@@ -12,6 +12,28 @@ Manage Indexes
    :depth: 1
    :class: singlecol
 
+Indexes are special data structures that support the efficient 
+execution of queries in MongoDB. Indexes store a portion of a
```
jeff-allen-mongo: Suggest making this last sentence (borrowed from MongoDB manual):

```
Indexes store a portion of a  collection's data in an easy-to-traverse form. The index stores
the value of a specific field or set of fields, ordered by the value of the field.
```

This way, when we do the bullets mentioned in my other comment, we give context for why the sort operations are pre-sorted.

source/indexes.txt
```diff
@@ -12,6 +12,28 @@ Manage Indexes
...
+
+To learn more about indexes, see :manual:`Indexes </indexes>`.
+
+.. admonition:: Considerations
+   :class: note
+
+   Indexes have some negative performance impact on write operations. 
+   For collections with high write-to-read ratio, indexes are expensive 
+   since each insert must also update any indexes. For a detailed list 
+   of considerations for Indexes, see
```
jeff-allen-mongo: I don't think this Indexes needs to be capitalized.

source/indexes.txt
```diff
@@ -12,6 +12,32 @@ Manage Indexes
...
+ordered by the value of the field.
+
+To improve query performance, build indexes on fields that appear often 
+in queries and for all operations that 
+:ref:`sort by a field <query-bar-sort>`.
+
+- Queries on an indexed field use the index to limit the number of 
+  documents that must be scanned to find matching documents.
+
+- Sort operations on an indexed field return documents pre-sorted
```
jeff-allen-mongo: So I think keeping the "can" is important because there are instances where the query might not be able to use the index for some unforeseen circumstance, so we don't want to claim it will always use it. The language used in the server docs is generally "x can use"

- Queries on an indexed field can use the index...

- Sort operations on an indexed field can return documents pre-sorted...

PR #232, opened 2020-03-06
(DOCSP-13484): 'Home' icon in Compass upper left corner has been replaced by the Cluster name

tahiya-mongo:
Jira: https://jira.mongodb.org/browse/DOCS-13484

Staging: https://docs-mongodbcom-staging.corp.mongodb.com/compass/tahiya.chowdhury/DOCS-13484/instance.html#compass-home


source/instance.txt
```diff
@@ -33,7 +33,7 @@ MongoDB instance to which Compass is connected, including:
    :alt: Compass Home View
 
 Once connected to a MongoDB instance, you can reach the Compass
-Home screen by clicking the :guilabel:`Home` icon in the upper left
+Home screen by clicking the cluster name in the upper left
```
jeff-allen-mongo: I'd suggest moving this descriptor above the image so users understand what we're highlighting.

PR #233, opened 2020-03-11
(DOCSP-8265) Update Compass import/export procedures

zach-carr:
[Ticket](https://jira.mongodb.org/browse/DOCSP-8265)
[Staged view](https://docs-mongodbcom-staging.corp.mongodb.com/compass/zach.carr/DOCSP-8265/import-export.html#)

Updated requirements, procedures and screenshots to match the import/export updates (from the ticket):

- we can handle imports from JSON files (JSON arrays), JSONlines is no longer a requirement
- export to JSON exports to JSON array
- it's possible to select the fields that are exported
- when importing from CSV, it's possible to select the type for each field

I removed the examples at the end of the page because the new UI breaks that information up into three dialog boxes making it harder to show clearly and concisely, and because I don't think it adds anything that the procedure doesn't cover clearly. 

I'm definitely open to adding the examples back in if there's disagreement - I saved the copy and the screenshots from the previous example.


source/includes/steps-import.yaml
```diff
@@ -19,25 +21,57 @@ content: |
...
 level: 4
 ref: choose-file-type
 content: |
-  Under :guilabel:`Select Input File Type`, select either JSON or CSV. 
+
+  Under :guilabel:`Select Input File Type`, select either 
+  :guilabel:`JSON` or :guilabel:`CSV`. 
+
+  If you are importing a CSV file, you may specify fields to import and 
+  the types of those fields under :guilabel:`Specify Fields and Types`:
```
jeff-allen-mongo: Does Compass automatically try to guess the type? Is there a default value? Might be worth clarifying what the "default" behavior is for data types.

source/includes/steps-import.yaml
```diff
@@ -19,25 +21,57 @@ content: |
...
 content: |
-  .. raw:: html
+  Under :guilabel:`Options`, select the import options for your use 
+  case. 
+  
+  If you are importing a CSV file, you may select how your data 
+  is delimited.
+
+  If checked, :guilabel:`Ignore empty strings` drops document fields 
+  that have an empty string value.
```
jeff-allen-mongo: An empty string value for any field? This is a little ambiguous.

source/includes/steps-import.yaml
```diff
@@ -19,25 +21,57 @@ content: |
...
+
+  If checked, :guilabel:`Stop on errors` prevents any data from being 
+  imported in the event of an error. If unchecked, data is inserted 
+  until an error is encountered and successful inserts are not rolled 
+  back.
```
jeff-allen-mongo: Does Compass stop the process entirely when it encounters an error, or will it process some documents, encounter an error, then try to process more documents? It might be worth the behavior when it continues after an error.

Also, do we want to provide an example of what might cause the process to error, if there's something common we can highlight to help users from being tripped up? These also might be questions for product.

source/includes/steps-import.yaml
```diff
@@ -19,25 +21,57 @@ content: |
...
+title: Choose the appropriate file type.
 level: 4
 ref: choose-file-type
 content: |
-  Under :guilabel:`Select Input File Type`, select either JSON or CSV. 
+
+  Under :guilabel:`Select Input File Type`, select either 
+  :guilabel:`JSON` or :guilabel:`CSV`. 
+
+  If you are importing a CSV file, you may specify fields to import and
```
jeff-allen-mongo: It was a little bit weird to see this content when I had selected the JSON tab in the previous section, but I see now that this content is not tabbed. Maybe it should be? I don't feel strongly either way, but something to think about since it didn't confuse me for a second.

source/includes/steps-import.yaml
```diff
@@ -19,25 +21,57 @@ content: |
...
+ref: import-options
 content: |
-  .. raw:: html
+  Under :guilabel:`Options`, select the import options for your use 
+  case. 
+  
+  If you are importing a CSV file, you may select how your data 
+  is delimited.
+
+  If checked, :guilabel:`Ignore empty strings` drops document fields
```
jeff-allen-mongo: It's not immediately clear that the next two settings apply to both JSON and CSV imports. Maybe tack on a sentence to the end of both like "You can set this setting for both JSON and CSV imports." or something like that.

source/includes/steps-import.yaml
```diff
@@ -19,25 +21,57 @@ content: |
...
+title: Configure import options.
 level: 4
-ref: click-file-location
+ref: import-options
 content: |
-  .. raw:: html
+  Under :guilabel:`Options`, select the import options for your use 
+  case. 
+  
+  If you are importing a CSV file, you may select how your data
```
jeff-allen-mongo: [optional] these might look nice as a bulleted list.

source/import-export.txt
```diff
@@ -49,20 +49,30 @@ that it is formatted correctly.
        name: JSON
```
jeff-allen-mongo: I can't comment on it directly, but at the top of the "Export" section on this page we have a blurb which states: "If you specify a filter in the query bar prior to export, Compass can optionally only export documents which match the specified query."

I don't think this is 100% true any more, since you can modify the filter prior to export. Can we modify that blurb to make it consistent with the new behavior?

source/includes/steps-export.yaml
```diff
@@ -7,36 +8,57 @@ content: |
...
-  .. figure:: /images/compass/export-data-dialog.png
-     :alt: Export data dialog
-     :figwidth: 600px
+  .. figure:: /images/compass/export-data-dialog-1.png
+     :alt: Export data dialog, step 1
+     :width: 60%
+
+  The export dialog initially displays the query entered in the query 
+  bar prior to export, if applicable. If no query was specified, this 
+  section displays a generic ``find`` operation.
```
jeff-allen-mongo: I don't love "generic" find operation. Let's be specific about what happens. "This section displays a ``find`` operation with no parameters, which returns all documents in the collection."

source/includes/steps-export.yaml
```diff
@@ -7,36 +8,57 @@ content: |
...
+
+  To ignore the query filter and export your entire collection, select 
+  :guilabel:`Export Full Collection`.
```
jeff-allen-mongo: Maybe direct them to click :guilabel:`Select Fields` ? Need some way to get from step 2 to 3.

source/includes/steps-export-filter.yaml
```diff
@@ -1,46 +1,70 @@
...
+title: Select document fields to include in your exported file.
 level: 4
-ref: click-file-type
+ref: select-export-fields
 content: |
-    Under :guilabel:`Select Output File Type`, select either JSON or CSV.
+  Only fields that are checked are included in the exported file.
+
+  You can add document fields to include with the :guilabel:`Add Field` 
+  button if the field you want to include is not automatically detected.
```
jeff-allen-mongo: You might want to explain the sampling behavior a bit here. Users might wonder why their field wasn't detected.

source/includes/steps-export-filter.yaml
```diff
@@ -1,46 +1,70 @@
...
-title: Under :guilabel:`Select File`, choose where to export the file.
+title: Choose the appropriate file type.
 level: 4
-ref: click-file-location
+ref: click-file-type
+content: |
+  Under :guilabel:`Select Export File Type`, select either 
+  :guilabel:`JSON` or :guilabel:`CSV`. If you select :guilabel:`JSON`,
+  your data is exported to the target file as a comma-separated 
+  array.
```
jeff-allen-mongo: Array of what? JSON objects?

source/includes/steps-import.yaml
zach-carr: Good call - defaults to strings.

source/includes/steps-import.yaml
zach-carr: It drops the empty field itself from the document, importing the rest of the document - I'll reword this.

source/includes/steps-import.yaml
zach-carr: They do

source/import-export.txt
zach-carr: Removed "...in the :ref:`query bar <compass-query-bar>` prior to export" since it's misleading. Let me know if you had more changes in mind.

source/includes/steps-export-filter.yaml
zach-carr: Yes, specifying

source/includes/steps-import.yaml
zach-carr: It stops completely in all cases - adding a sentence about that.

Just guessing, but the most common error may be 'duplicate keys' if you try to reinsert a doc that's already been imported, or possibly something like "invalid token"  if it's not formatted properly. I'll call this out for tech review.

source/includes/steps-import.yaml
jeff-allen-mongo: Oh sorry, I think this was a case of me misreading. I read it as "drops documents" that have an empty string value. Feel free to reword but I think what you have might be ok.

source/includes/steps-export-filter.yaml
zach-carr: I don't think this has to do with sampling because you're exporting a full collection or a partial collection defined by a filter. I think Compass may leave out fields that don't appear on some % of documents. I'll call this out in tech review as well and add some context.

source/includes/steps-import.yaml
zach-carr: I agree that it's weird to see this at all, but will try to improve it along the lines of your next comment: differentiating the CSV-only and JSON + CSV options. The distinction is made in one of the YAML procedure files, so to tab it out I think I would need to move that content to the import-export.txt file. That's definitely doable if we decide it's worth it, but for the sake of the existing structure I'll leave as-is for now.

zach-carr:
Hi @mmarcon, this is ready for tech review when you get some time.

From my discussion with Jeff, I wanted to ask:
1. Are there any common errors, importing or exporting, that trip users up we should call out or make an example of?
2. When exporting, why might you need to add a field?

I also wanted to note that I've been getting this error on Version 1.21.0-beta.1 (1.21.0-beta.1) when exporting collections: 

Cursor session id (aa30f2bf-c221-48b6-847f-f81d7f82187d - ddqot68vLpBUPnOuqFj2cdClv/dv/vF9ZVSwknYHEsE=) is not the same as the operation context's session id (none)

It successfully exports the first 101 documents then fails. I'll test again tomorrow, I may be doing something wrong.

mmarcon:
@zach-carr thank you for putting this together.

1. Not sure. In my tests, I have not run into any errors so far but some early adopters have seen some errors (`Cannot run getMore on cursor 1998783333627462641, which was created in session a7a167f1-dac6-4fd4-9000-cfd2b3270274 - 47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=, without an lsid`). Give us some time to get some more user feedback so we can see if there are some recurring issues that we want to document.
2. The pre-populated list of fields is generated by sampling the collection. When we sample and the documents don't have all the same schema, we may miss some fields. The user may know of the existence of a field that wasn't detected by the sampling so we give them the possibility to add it.

@imlucas FYI

mmarcon:
@zach-carr just got the same error, I am going to create a ticket.

source/includes/steps-export-filter.yaml
zach-carr: My mistake, it does detect fields by sampling. Will add some context like you suggested.

zach-carr:
Thanks for those answers and creating the error ticket @mmarcon.

@jeff-allen-mongo, ready for another look when you have time.

source/includes/steps-import.yaml
jeff-allen-mongo: Ok we can leave as-is.

source/import-export.txt
```diff
@@ -49,20 +49,30 @@ that it is formatted correctly.
...
 
                { "type": "home", "number": "212-555-1234" }
                { "type": "cell", "number": "646-555-4567" }
                { "type": "office", "number": "202-555-0182"}
 
+            The following comma-separated ``.json`` array file is also formatted correctly:
+
+            .. code-block:: javascript
+
+               [{ "type": "home", "number": "212-555-1234" }, { "type": "cell", "number": "646-555-4567" }, { "type": "office", "number": "202-555-0182"}]
```
jeff-allen-mongo: Does this format only work if all of the items are on the same line?

source/import-export.txt
```diff
@@ -109,8 +120,7 @@ Export Data from a Collection
 
 |compass| can export data from a collection as either a
 **JSON** or **CSV** file. If you specify a
-:ref:`filter <query-bar-filter>` in the
-:ref:`query bar <compass-query-bar>` prior to export, |compass-short|
+:ref:`filter <query-bar-filter>`, |compass-short|
```
jeff-allen-mongo: I'd reword this a bit to say: "Compass only exports documents which match the specified query." With the new flow, you're explicitly specifying an export filter, so I think we can be more direct.

source/includes/steps-export-filter.yaml
```diff
@@ -1,46 +1,76 @@
...
+  You can add document fields to include with the :guilabel:`Add Field` 
+  button if the field you want to include is not automatically detected.
+
+  .. figure:: /images/compass/export-data-dialog-2.png
+     :alt: Export data dialog, step 2
+     :width: 60%
+
+  .. note::
+     
+     |compass-short| samples your collection to pre-populate a list of fields. Uncommon fields may not be automatically detected.
```
jeff-allen-mongo: It's not super clear what "uncommon" means in this context. Maybe say "Fields which only appear in a small percentage of documents may not be automatically detected."

Also need a wrap here.

source/includes/steps-export-filter.yaml
```diff
@@ -1,46 +1,76 @@
...
-
   .. figure:: /images/compass/export-data-filter-option-select.png
      :alt: Export data select menu
 
   |compass-short| displays the following dialog:
 
   .. figure:: /images/compass/export-data-dialog-filter.png
      :alt: Export data dialog
-     :figwidth: 600px
+     :width: 60%
```
jeff-allen-mongo: 1. Can we make this a bit bigger? Maybe 80% width? In staging it's a little blurry.

2. When the image says "Recommended" what does that mean?

source/includes/steps-import.yaml
```diff
@@ -19,25 +21,61 @@ content: |
...
+ref: import-options
 content: |
-  .. raw:: html
+  Under :guilabel:`Options`, configure the import options for your use 
+  case.
+  
+  If you are importing a CSV file, you may select how your data 
+  is delimited.
+
+  For both JSON and CSV file imports, you can toggle :guilabel:`Ignore empty strings` and :guilabel:`Stop on errors`:
```
jeff-allen-mongo: Wrap here

source/includes/steps-import.yaml
```diff
@@ -19,25 +21,61 @@ content: |
...
 level: 4
 ref: choose-file-type
 content: |
-  Under :guilabel:`Select Input File Type`, select either JSON or CSV. 
+
+  Under :guilabel:`Select Input File Type`, select either 
+  :guilabel:`JSON` or :guilabel:`CSV`. 
+
+  If you are importing a CSV file, you may specify fields to import and 
+  the types of those fields under :guilabel:`Specify Fields and Types`. By default, |compass-short| imports every field as a string.
```
jeff-allen-mongo: Wrap here. Also, I think we can tighten up the last sentence a bit: "The default data type for all fields is string".

source/includes/steps-export-filter.yaml
zach-carr: 1. Upped all the screenshots to 80% to keep them consistent.

2. I'm not sure. It does always seem to say exporting with a filter is (Recommended), even when the filter is a blank find and should be the same as exporting all docs. @mmarcon, do you know why Compass recommends exporting with a filter?

source/import-export.txt
zach-carr: No, it will ignore line breaks. You could format like this as well:
[
   { ...someJSON },
   { ...so on },
]
I put the JSON example array on the same line to illustrate the difference between formats. I'll add a note that the JSON array format will ignore line breaks, but am open to any format for this. Let me know if you think it's better to show on separate lines.

zach-carr:
Thanks @jeff-allen-mongo, made those changes.

Let me know what you think about the JSON array example and line "|compass-short| ignores line breaks in JSON arrays."

source/import-export.txt
```diff
@@ -49,20 +49,33 @@ that it is formatted correctly.
...
                { "type": "office", "number": "202-555-0182"}
 
+            The following comma-separated ``.json`` array file is also 
+            formatted correctly:
+
+            .. code-block:: javascript
+
+               [{ "type": "home", "number": "212-555-1234" }, { "type": "cell", "number": "646-555-4567" }, { "type": "office", "number": "202-555-0182"}]
+
+            |compass-short| ignores line breaks in JSON arrays.
```
jeff-allen-mongo: [super nit] Can we make the following paragraph |compass-short| instead of |compass| to keep them parallel?

jeff-allen-mongo:
Btw now that I see the images enlarged I'm not sure if they actually look better. Up to you if you want to keep them as-is or revert to a smaller size.

source/import-export.txt
zach-carr: Done, and went with a happy medium of 70% for the images.

PR #235, opened 2020-03-16
Fixed preferences path on Windows

mmarcon:
Just tried this and it looks like `%APPDATA%`
already points to the `Roaming` directory.


jeff-allen-mongo:
@durran can you confirm that this change is accurate?

PR #243, opened 2020-03-30
(DOCSP-9614) Document export schema feature

zach-carr:
[Ticket](https://jira.mongodb.org/browse/DOCSP-9614)
[Staged](https://docs-mongodbcom-staging.corp.mongodb.com/compass/zach.carr/DOCSP-9614/schema.html#export-your-schema)


source/schema.txt
```diff
@@ -338,3 +338,103 @@ To delete a location filter from the map:
...
+
+- ``fields`` is an array of metadata for each analyzed field, including 
+  the field's ``name``, ``types``, and within ``types``, an array of 
+  the actual values analyzed for each type.
+
+- ``count`` is the number of documents :ref:`sampled <sampling>` from 
+  the collection.
+
+The following contrived example shows a schema for a collection 
+of 3 documents with only ``_id`` and ``user`` fields:
```
jeff-allen-mongo: Just curious why we're using such a simple example? It might be more interesting to show a schema with embedded objects, and possibly arrays with mixed types.

Also, if the example only has three documents, we might as well provide them to the user in case they want to import / play around with them themselves.

zach-carr: Good ideas, making it a bit more complex to show embedded data and adding the documents to import.

The simple example was to keep the schema output simple so that users can easily see how documents map to it. I think with anything like a real collection the schema is pretty massive and harder to understand without really scrutinizing.

source/schema.txt
```diff
@@ -338,3 +338,103 @@ To delete a location filter from the map:
...
+
+#. From the dropdown, click :guilabel:`Share Schema as JSON`.
+
+.. figure:: /images/compass/share-schema-json.png
+   :alt: Image showing Collection dropdown
+
+Your schema is copied to your clipboard as a |json| object with 
+properties ``fields`` and ``count``:
+
+- ``fields`` is an array of metadata for each analyzed field, including
```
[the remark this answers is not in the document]
zach-carr: Will do - I like the table

source/schema.txt
```diff
@@ -338,3 +338,424 @@ To delete a location filter from the map:
 .. |query-build-ref| replace:: :ref:`build-query`
 
 .. |query-build-type| replace:: :ref:`build-query`
+
+Export Your Schema
```
jeff-allen-mongo, line 342: I should've mentioned this earlier, but we probably want this on a sub-page of the main Schema Analysis page. This page was long already and adding this section certainly isn't helping the cause.

source/schema.txt
```diff
@@ -338,3 +338,424 @@ To delete a location filter from the map:
...
+------------------
+
+You can export your schema after analyzing it. This is useful for 
+sharing your schema and comparing schemas across collections.
```
jeff-allen-mongo, line 346: [note] let's call attention to this in external review to see if there are any other specific use cases we want to highlight.

source/schema.txt
```diff
@@ -338,3 +338,424 @@ To delete a location filter from the map:
...
+
+.. figure:: /images/compass/share-schema-json.png
+   :alt: Image showing Collection dropdown
+
+Your schema is copied to your clipboard as a |json| object.
+
+Schema Object Properties
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Schema objects have properties ``count`` and ``fields``:
```
jeff-allen-mongo, line 373: [nit] I think this is easier to read if we say Schema objects have count and fields properties:

source/schema.txt
```diff
@@ -338,3 +338,424 @@ To delete a location filter from the map:
...
+
+- ``count`` is an integer that represents the number of documents 
+  :ref:`sampled <sampling>` from the collection to generate the schema.
+
+- ``fields`` is an array of ``n`` metadata objects that represent each
```
jeff-allen-mongo, line 378: I don't think we need `n` here. I think when we use `n` below it should be clear what we're talking about since it's an array.

source/schema.txt
```diff
@@ -338,3 +338,424 @@ To delete a location filter from the map:
...
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Schema objects have properties ``count`` and ``fields``:
+
+- ``count`` is an integer that represents the number of documents 
+  :ref:`sampled <sampling>` from the collection to generate the schema.
+
+- ``fields`` is an array of ``n`` metadata objects that represent each 
+  field in the documents analyzed during sampling. Each element in the 
+  ``fields`` array contains:
```
jeff-allen-mongo, line 380: [nit] contains the following fields:

source/schema.txt
```diff
@@ -338,3 +338,424 @@ To delete a location filter from the map:
...
+    - String
+    - Name of ``fields[n]``, e.g. ``_id``.
+
+  * - ``path``
+    - String
+    - Path to ``fields[n]`` within a document.
+    
+  * - ``count``
+    - Integer
+    - The number of documents that ``fields[n]`` appears on.
```
jeff-allen-mongo, line 400: "on" or "in"? Fields are typically "in" documents. Also, to avoid ending with a preposition: Number of documents in which fields[n] appears.

source/schema.txt
```diff
@@ -338,3 +338,424 @@ To delete a location filter from the map:
...
+
+  * - ``types``
+    - Array
+    - List of ``n`` metadata objects that represent each data type that 
+      appears in ``fields[n]``.
```
jeff-allen-mongo, line 405: Consideration: I know we've gone back and forth on this but I'm wondering if "appears in the current `fields` array element" is maybe clearer. Up to you.

source/schema.txt
```diff
@@ -338,3 +338,424 @@ To delete a location filter from the map:
...
+    - String
+    - Data type of ``types[n]``.
+  
+  * - ``types[n].bsonType``
+    - String
+    - |bson| type of ``types[n]``.
+
+  * - ``types[n].path``
+    - String
+    - Path to ``name`` of ``fields[n]`` within a document.
```
jeff-allen-mongo, line 417: I think using `name` here is a little confusing because we have both `types[n].name` and `name`, so it's unclear which one we mean. What about: "Path to `fields[n]` where `fields[n]` has this data type."

source/schema.txt
```diff
@@ -338,3 +338,424 @@ To delete a location filter from the map:
...
+
+  * - ``types[n].count``
+    - Integer
+    - Number of times ``types[n]`` appears in ``fields[n]``.
```
jeff-allen-mongo, line 421: Reading this again, I think "this data type" is clearer than `types[n]`.

source/schema.txt
```diff
@@ -338,3 +338,424 @@ To delete a location filter from the map:
...
+      ``fields[n]`` is not a nested value.
+  
+  * - ``types[n].probability``
+    - Number
+    - Probability that the value of ``field[n]`` is ``types[n]`` in a 
+      random document.
+
+  * - ``types[n].unique``
+    - Integer
+    - Number of unique values of ``types[n]`` that appear in
```
jeff-allen-mongo, line 440: I think replace `types[n]` with "this data type"

source/schema.txt
```diff
@@ -338,3 +338,424 @@ To delete a location filter from the map:
...
+      appear in ``fields[n]`` and match ``types[n]``.
+
+  * - ``types[n].total_count``
+    - Integer
+    - Number of items nested in ``field[n]``. 0 when 
+      ``fields[n]`` is not a nested value.
+  
+  * - ``types[n].probability``
+    - Number
+    - Probability that the value of ``field[n]`` is ``types[n]`` in a
```
jeff-allen-mongo, line 435: I think replace types[n] with "this data type"

source/schema.txt
```diff
@@ -338,3 +338,424 @@ To delete a location filter from the map:
...
+    - Number of times ``types[n]`` appears in ``fields[n]``.
+  
+  * - ``types[n].values``
+    - Array
+    - List of the actual sampled values that 
+      appear in ``fields[n]`` and match ``types[n]``.
+
+  * - ``types[n].total_count``
+    - Integer
+    - Number of items nested in ``field[n]``. 0 when
```
jeff-allen-mongo, line 430: I think the sentence sentence is more clearly read as "When `fields[n]` is not an embedded value, `types[n].total_count` is 0.

source/schema.txt
```diff
@@ -338,3 +338,424 @@ To delete a location filter from the map:
...
+    - Path to ``name`` of ``fields[n]`` within a document.
+
+  * - ``types[n].count``
+    - Integer
+    - Number of times ``types[n]`` appears in ``fields[n]``.
+  
+  * - ``types[n].values``
+    - Array
+    - List of the actual sampled values that 
+      appear in ``fields[n]`` and match ``types[n]``.
```
jeff-allen-mongo, line 426: I think replace types[n] with "this data type"

source/schema.txt
```diff
@@ -338,3 +338,424 @@ To delete a location filter from the map:
...
+      random document.
+
+  * - ``types[n].unique``
+    - Integer
+    - Number of unique values of ``types[n]`` that appear in 
+      ``fields[n]``.
+
+  * - ``types[n].has_duplicates``
+    - Boolean
+    - ``true`` if a single value of ``types[n]`` appears multiple times
```
jeff-allen-mongo, line 445: I think replace types[n] with "this data type"

source/schema.txt
```diff
@@ -338,3 +338,424 @@ To delete a location filter from the map:
...
+      ``fields[n]``.
+
+  * - ``types[n].has_duplicates``
+    - Boolean
+    - ``true`` if a single value of ``types[n]`` appears multiple times 
+      in ``fields[n]``. Otherwise ``false``.
+
+  * - ``types[n].lengths``
+    - Array
+    - If ``types[n]`` is an array, an array of integers representing
```
jeff-allen-mongo, line 450: I think replace types[n] with "this data type"

source/schema.txt
```diff
@@ -338,3 +338,424 @@ To delete a location filter from the map:
...
+
+  * - ``types[n].lengths``
+    - Array
+    - If ``types[n]`` is an array, an array of integers representing 
+      the lengths of arrays found in ``fields[n]``. Not present for 
+      other data types.
+
+  * - ``types[n].average_length``
+    - Number
+    - If ``types[n]`` is an array, the average length of arrays in
```
jeff-allen-mongo, line 456: I think replace types[n] with "this data type"

source/schema.txt
```diff
@@ -338,3 +338,424 @@ To delete a location filter from the map:
...
+
+  * - ``types[n].average_length``
+    - Number
+    - If ``types[n]`` is an array, the average length of arrays in 
+      ``fields[n]`` across sampled documents. Not present for 
+      other data types.
+
+  * - ``total_count``
+    - Integer
+    - Number of documents sampled from the collection.
```
jeff-allen-mongo, line 462: That contain `fields[n]`? or total number overall?

source/schema.txt
```diff
@@ -338,3 +338,424 @@ To delete a location filter from the map:
...
+      ``fields[n]``.
+
+  * - ``has_duplicates``
+    - Boolean
+    - ``true`` if a single value appears multiple times in 
+      ``fields[n]``. Otherwise ``false``.
+
+  * - ``probability``
+    - Number
+    - Probability that a random document has ``fields[n]``.
```
jeff-allen-mongo, line 476: [nit] has -> ccontains

source/schema.txt
zach-carr: Went with: "Path to ``fields[n]`` within a document." which is consistent with the use of ``fields[n]`` as a stand-in for the document field, like _id. That also makes it the same as the description for ``path``, which is correct - it is always the same value as the parent's path.

source/schema.txt
zach-carr: Agreed

source/schema.txt
zach-carr: Noting that I will call this out for tech review - this is the entry I'm not totally sure about.

source/schema.txt
zach-carr: Total overall - I just double checked on a sample airbnb collection. I'll call out for tech review just in case.

source/schema.txt
zach-carr: I agree this has to be changed again across the board - ``fields[n]`` is an object, so using it to refer to the field we're trying to talk about will be confusing to someone. I don't think we can say "current fields array element" for the same reason. It's more like "the field in your documents that ``fields[n]`` describes." Brainstorming more.

zach-carr: I might have overthought this. Going with "this field," I think it's intuitive and reads well.

Changing the table heading to "property" instead of "field name" to cut down on using "field."

zach-carr:
Thanks @jeff-allen-mongo, this is ready for another look.

[New stage](https://docs-mongodbcom-staging.corp.mongodb.com/compass/zach.carr/DOCSP-9614/schema/export.html)

I decided to keep the tables together because I felt like it was more intuitive with the use of "this field". Separating them looks good but then I think we have an even more difficult naming problem. Let me know what you think.

zach-carr:
Hi @mmarcon, would you mind reviewing these docs when you have some time? Thank you.

One question to call out: Is the description of ``types[n].total_count`` accurate? As best I could tell it is 0 for single values, the # of elements for array values, and is embedded further for object values, ultimately being 1 for each property.

mmarcon:
LGTM but I don't know the answer to that question. @durran can you please comment?

You probably want to do that in a separate PR, but the map screenshot on this section is outdated: https://docs-mongodbcom-staging.corp.mongodb.com/compass/zach.carr/DOCSP-9614/schema.html#geojson-and-latitude-longitude-arrays

zach-carr:
Thanks Max. I created [DOCSP-9892](https://jira.mongodb.org/browse/DOCSP-9892) to update the map screenshot - will get to that soon. 

@durran When you have a moment, I have a question about the exported schema's ``types.total_count`` property that I'm hoping you can answer. Thanks in advance!

Q:
Is the description of types[n].total_count accurate? As best I could tell it is 0 for single values, the # of elements for array values, and is embedded further for object values, ultimately being 1 for each property.

durran:
@zach-carr Basically the only time it's really calculated is for array values, which is the number of elements. The numbers for any other type are just accidental and have no meaning.

zach-carr:
Thanks @durran!

@mmarcon when you have a moment can I get your final approval?

PR #252, opened 2020-04-20
(DOCSP-9972): Landing page for vsce

jeff-allen-mongo:
Staged: [VS Code Extension](https://docs-mongodbcom-staging.corp.mongodb.com/compass/docsworker/DOCSP-9972/vs-code-extension.html)


jeff-allen-mongo:
Hey @mmarcon, would you mind taking a look at this initial draft of the VS Code Extension landing page? I've also included a shell (titles only) for the other pages we'll be adding, so feel free to advise if you think there's a better way to structure this content.

I wasn't sure if there was any other functionality or maybe use cases we wanted to highlight on the landing page? There's not a ton there right now (which is why I tried to fill up some space with a screenshot), so definitely open to ideas if you think it'd be good to add any more info.

Thanks!
Jeff

mmarcon:
Here are my thoughts (@sgrinfy feel free to add more).

### General
The product name is _MongoDB for VS Code_

### Navigation (not sure if this is in scope for this PR)
* MONGODB COMPASS should probably become MONGODB DEVELOPER TOOLS
* Can we have a Compass section at the same level of VS Code with the Compass items nested under that? 

### Landing page
* Screenshot: I'd unfold the Schema section of the tree-view
* I'd add to the highlights [Quick access to the MongoDB Shell](https://github.com/mongodb-js/vscode#quick-access-to-the-mongodb-shell)
* _Write and execute aggregation pipelines on your data_ and _Write queries and perform CRUD operations on your data_: how do you feel about having those as sub-bullets of an item that is something like "Playgrounds: quickly prototype queries and run MongoDB commands

### Sections
Sections look good but I wonder if somehow we want to call out that Playgrounds are for any MongoDB command and that one could run any JS, as subset of which are CRUD operations and aggregations.

jeff-allen-mongo:
Thanks for the feedback @mmarcon. Some notes:

Updated throughout with the name "MongoDB for VS Code"

Re Navigation: I think this is a good approach but I have a few questions:
1. How do we want to handle the version dropdown? Right now, we document Compass stable and Compass beta. We could either keep the version dropdown as-is and have it affect the Compass docs, but then I don't think it's clear what it's doing. We could potentially it entirely and only document the Stable version of Compass.
2. Would we want to change the landing URL for these docs? Right now it's `docs.mongodb.com/compass`. We could possibly update to something like `docs.mongodb.com/developer-tools` and have `docs.mongodb.com/compass` redirect to the new URL.

Re Landing Page:

- I re-used the screenshot from the code repo but cropped out the "dev host" menu bar at the top. I like the idea of showing something a bit "busier" in the screenshot than a simple document.

- Added the quick access to the shell highlight.

- Reformatted slightly to better call out Playgrounds. Now we have a top-level page for Playgrounds where we can explain what they are, and sub-pages for the specific use cases.

mmarcon:
Thank you for making the changes. Would it be ok to talk about 1 and 2 in tomorrow's meeting?

jeff-allen-mongo:
Sounds good to me 👍

jeff-allen-mongo:
Closing this out since VSCE is now its own repo.

PR #258, opened 2020-05-05
(DOCSP-10067) Troubleshooting for schema analysis error

zach-carr:
[Ticket](https://jira.mongodb.org/browse/DOCSP-10067)
[Staged section](https://docs-mongodbcom-staging.corp.mongodb.com/compass/zach.carr/DOCSP-10067/schema.html#troubleshooting)

This will need to be ported to the beta branch as well.


source/schema.txt
```diff
@@ -335,6 +335,28 @@ To delete a location filter from the map:
 
 .. |query-build-type| replace:: :ref:`build-query`
 
+Troubleshooting
+---------------
+
+If the analysis of your schema times out, it might be because the 
+collection you are analyzing is very large and the operation is stopped
```
nathan-contino-mongo, line 342: "is stopped" sounds a little strange to me here. I think it could sound better if we reword -- how about something like:

```suggestion
collection you are analyzing is very large, causing MongoDB to stop the operation 
```

source/schema.txt
```diff
@@ -335,6 +335,28 @@ To delete a location filter from the map:
...
+before the analysis is complete. Increase the value of ``maxTimeMS`` 
+to allow the operation time to complete.
+
+To increase the value of ``maxTimeMS``:
+
+1. In the query bar, expand :guilabel:`Options`.
+
+   .. figure:: /images/compass/schema-query-options.png
+      :alt: The Options button is on the right side of the query bar, 
+            next to the green Analyze button.
```
nathan-contino-mongo, line 352: I think we tend to `guilabel` button text in DevEd -- if you do the same, you might want to make this change!

```suggestion
            next to the green :guilabel:`Analyze` button.
```

source/schema.txt
```diff
@@ -335,6 +335,28 @@ To delete a location filter from the map:
...
+
+2. Increase the value of ``maxTimeMS`` to accommodate your collection. 
+   The value you should use depends on your collection.
```
nathan-contino-mongo, line 355: I feel like this is really vague, but I also suspect there's no good general guidance to give on this subject so I don't really know if it can be improved at all. Maybe give some kind of ballpark estimate, like "``maxTimeMS`` defaults to 500ms, but large collections can take multiple seconds to analyze"?

source/schema.txt
zach-carr: Agreed, took that suggestion

source/schema.txt
zach-carr: Normally we would but this is an extension of that alt text (the line wrapping is out of whack) and I don't think directives work there

source/schema.txt
zach-carr: Agreed, that was too vague - going with something similar to your suggestion. I'll call this part out in tech review to see if there's a more concrete recommendation to make.

zach-carr:
Thanks Nate!

@mmarcon, would you please review this schema analysis Troubleshooting section when you have a moment? I want to call out the maxTimeMS suggestion in step 2 - is there a more specific recommendation to make?

[Staged section](https://docs-mongodbcom-staging.corp.mongodb.com/compass/zach.carr/DOCSP-10067/schema.html#troubleshooting)

PR #263, opened 2020-06-02
(DOCSP-10513): Run Compass as a different user for Kerb Auth

jeff-allen-mongo:
There's some technical details in the ticket here: [DOCSP-10513](https://jira.mongodb.org/browse/DOCSP-10513)

Staging: [Connect to MongoDB](https://docs-mongodbcom-staging.corp.mongodb.com/compass/docsworker/DOCSP-10513/connect.html#connect)

Fill in Individual Fields Tab > Scroll down to Kerberos Tab in the Authentication row.


source/includes/steps-starting-compass-individual-fields.yaml
```diff
@@ -156,6 +161,43 @@ content: |
...
+
+                   To authenticate as a different Kerberos user, you can
+                   run |compass| as the desired user, and instead
+                   specify the :guilabel:`Principal` for that user.
+
+                   .. example::
+
+                      To run |compass| as a user named ``admin``:
+
+                      .. code-block:: sh
```
nathan-contino-mongo, line 179: The highlighting for this renders a little weirdly for me (specifically around that backtick, which it seems to interpret as an escape character. Perhaps you could change this to:
```suggestion
                      .. code-block:: text
```
Since the highlighting is neither useful nor important here?

jeff-allen-mongo: Agreed. Updating.

PR #267, opened 2020-06-22
(DOCSP-10869) Note on Automatic Creation of Object ID

jocelyn-mendez:
Added note stating automatic creation of an object ID when one is not provided during document insertion.
Jira: https://jira.mongodb.org/browse/DOCSP-10869
Staging: https://docs-mongodbcom-staging.corp.mongodb.com/compass/docsworker/DOCSP-10869/index.html

### DIFF
--- a/source/documents/insert.txt
+++ b/source/documents/insert.txt
@@ -81,7 +81,7 @@ To insert documents into your collection:
 
             .. note::
               
-              ..include:: /includes/fact-automatic-objectid.rst
+              .. include:: /includes/fact-automatic-objectid.rst
               
 
          4. Click :guilabel:`Insert`.
@@ -99,7 +99,7 @@ To insert documents into your collection:
 
             .. note::
               
-              ..include:: /includes/fact-automatic-objectid.rst
+              .. include:: /includes/fact-automatic-objectid.rst
 
          Add New Fields
          ~~~~~~~~~~~~~~
--- a/source/documents/insert.txt
+++ b/source/documents/insert.txt
@@ -79,6 +79,10 @@ To insert documents into your collection:
                     { "_id" : 8751, "title" : "The Banquet", "author" : "Dante", "copies" : 2 }
                   ]
 
+            .. note::
+
+              If an object ID is not provided by user, Compass will automatically generate an object ID.
+
          4. Click :guilabel:`Insert`.
 
       .. tab:: Field-by-Field Editor
--- a/source/documents/insert.txt
+++ b/source/documents/insert.txt
@@ -80,8 +80,9 @@ To insert documents into your collection:
                   ]
 
             .. note::
-
-              If an object ID is not provided by user, Compass will automatically generate an object ID.
+              
+              ..include:: /includes/fact-automatic-objectid.rst
+              
 
          4. Click :guilabel:`Insert`.
 
@@ -98,7 +99,7 @@ To insert documents into your collection:
 
             .. note::
               
-              If an object ID is not provided by user, Compass will automatically generate an object ID.
+              ..include:: /includes/fact-automatic-objectid.rst
 
          Add New Fields
          ~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


source/documents/insert.txt
```diff
@@ -79,6 +79,10 @@ To insert documents into your collection:
                     { "_id" : 8751, "title" : "The Banquet", "author" : "Dante", "copies" : 2 }
                   ]
 
+            .. note::
+
+              If an object ID is not provided by user, Compass will automatically generate an object ID.
```
jeff-allen-mongo, line 84: [leaving a few suggestions to help improve tone and be more direct]

Typically, we want to use active voice wherever possible. So, instead of saying 

"If an object ID is not provided by the user..." 

we can say

"If you do not provide an ObjectId..."

Second person is encouraged where it fits.

jeff-allen-mongo, line 84: We default to present tense when possible, so instead of

"Compass will automatically generate an objectId"

we can say

"Compass automatically generates an ObjectId"

jeff-allen-mongo, line 84: It may be helpful to link the word "ObjectId" to the [corresponding page in the manual](https://docs.mongodb.com/manual/reference/bson-types/#objectid)

To do so, you can do:

```
:manual:`ObjectId </reference/bson-types/#objectid>`
```

I can explain this in more detail via Slack, but you can look at the `extlinks` in the `conf.py` file.

jeff-allen-mongo, line 84: "object ID" should be written as "ObjectId". See: https://docs.mongodb.com/manual/reference/bson-types/#objectid for reference

jeff-allen-mongo, line 84: We actually have a replacement we use for the product name "Compass". You can replace "Compass" here with `|compass-short|`. This is the type of replacement Zach was mentioning in standup today, if you recall.

We do this so that if for whatever reason the product name changes, it's easy for us to change throughout the docs just by modifying the replacement.

If you want to take a look, all of the replacements you can use are in the `conf.py` file of the repo.

source/documents/insert.txt
```diff
@@ -92,6 +96,10 @@ To insert documents into your collection:
             .. figure:: /images/compass/insert-document2.png
               :figwidth: 590px
 
+            .. note::
```
jeff-allen-mongo, line 99: Thanks for remembering to do both tabs! Reminder to move the copy for this to a singe include file as discussed via Zoom.

source/documents/insert.txt
jeff-allen-mongo: [note] I'd only do this for the _first_ occurrence of "ObjectId". One link is enough

source/includes/fact-automatic-objectid.rst
```diff
@@ -0,0 +1 @@
+If you do not provide an :manual:`ObjectId </reference/bson-types/#objectid>, |compass-short| automatically generates an ObjectId.
```
jeff-allen-mongo, line 1: [small update] I would add:

"If you do not provide an ObjectId _in your document_ ... "

jeff-allen-mongo, line 1: Please wrap to 72 characters as noted in Slack

source/includes/fact-automatic-objectid.rst
```diff
@@ -0,0 +1,3 @@
+If you do not provide an :manual:`ObjectId
+</reference/bson-types/#objectid> in your document, |compass-short|
```
jeff-allen-mongo, line 2: Hey @jocelyn-mendez, check this out in staging; it's not rendering correctly. You're missing a closing backtick. See this suggestion:

```suggestion
</reference/bson-types/#objectid>` in your document, |compass-short|
```

jocelyn-mendez: @jeff-allen-mongo Thanks just noticed it!

PR #270, opened 2020-06-29
DOCSP-10756) Added step noting directions for Compass to trust macOS

jocelyn-mendez:
Added step number 7 to note how to allow macOS to trust Compass in order to fix/prevent security error.

Jira: https://jira.mongodb.org/browse/DOCSP-10756#add-comment

Staging: https://docs-mongodbcom-staging.corp.mongodb.com/compass/docsworker/DOCSP-10756/index.html

### DIFF
--- a/source/install.txt
+++ b/source/install.txt
@@ -155,10 +155,10 @@ Install Compass
             permissions to run. You may be prompted to enter your
             system password before launching |compass-short|.
          
-         #. Allow macOS to trust |compass-short|. If you receive a
-         security error when starting |compass-short| indicating that the
-         developer could not be identified or verified, perform the
-         following actions:
+         7. Allow macOS to trust |compass-short|. If you receive a
+         security error when starting |compass-short| indicating that
+         the developer could not be identified or verified, perform the
+         following actions to allow |compass-short| to run:
             
             a. Open *System Preferences*.
    
--- a/source/install.txt
+++ b/source/install.txt
@@ -155,10 +155,11 @@ Install Compass
             permissions to run. You may be prompted to enter your
             system password before launching |compass-short|.
          
+         
          7. Allow macOS to trust |compass-short|. If you receive a
-         security error when starting |compass-short| indicating that
-         the developer could not be identified or verified, perform the
-         following actions to allow |compass-short| to run:
+            security error when starting |compass-short| indicating that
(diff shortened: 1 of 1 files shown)


source/install.txt
```diff
@@ -154,6 +154,23 @@ Install Compass
             to modify your system settings to grant |compass-short|
             permissions to run. You may be prompted to enter your
             system password before launching |compass-short|.
+         
+         
+         7. Allow macOS to trust |compass-short|. If you receive a
```
jeff-allen-mongo, line 159: Also, this is slightly out of scope for this specific ticket, but the current screenshot of the "Downloaded from the Internet" dialog is dated 2015, which may give off the impression that these docs haven't been updated for 5 years (not true, but it may come off that way).

I'd suggest as part of this change that we replace that step with something that specifically describes the dialog box and remove the screenshot. For example:

```
#. When you open |compass| for the first time, you may receive a notice stating
   that it is an application downloaded from the internet, requiring you to confirm
   you want to open it. Click :guilabel:`Open` to continue and launch |compass-short|.
```

source/install.txt
```diff
@@ -154,6 +154,23 @@ Install Compass
...
+            security error when starting |compass-short| indicating that
+            the developer could not be identified or verified, perform the
+            following actions to allow |compass-short| to run:
+            
+            a. Open *System Preferences*.
+   
+            #. Select the *Security and Privacy* pane.
+
+            #. Under the *General* tab, click the button to the right of the
```
jeff-allen-mongo, line 168: After this step, I think they may need to re-run Compass in this case, so possibly add a step:

`d. If necessary, re-open |compass-short|.`

jocelyn-mendez:
Thanks for the feedback @jeff-allen-mongo! I went through and made some changes, let me know what you think when you get a chance.

jocelyn-mendez:
Hi @mmarcon, the install procedure has been updated and we are wondering if the Upgrade procedure should be updated as well? Please let me know your thoughts when you get a chance, thanks!

mmarcon:
LGTM.

Regarding your question on the upgrade procedure, are you referring to this page? https://docs.mongodb.com/compass/master/upgrade/

It might be worth mentioning that on Catalina and later users might have to explicitly tell macOS to allow Compass to run and maybe link to the section on the installation page.

PR #276, opened 2020-07-08
(DOCSP-11049) Added note on macOS permission for Compass to run

jocelyn-mendez:
Added note that macOS Catalina, and later versions, may require permission to be granted to allow Compass to run. 

jira: https://jira.mongodb.org/browse/DOCSP-11049

staging: https://docs-mongodbcom-staging.corp.mongodb.com/compass/jocelynmendez/DOCSP-11049/index.html

### DIFF
--- a/source/upgrade.txt
+++ b/source/upgrade.txt
@@ -104,7 +104,7 @@ Procedure
    If you are running macOS Catalina, or later, you may have to modify
    your system settings to grant |compass-short| permission to run. For
    more information on how to grant access permission to
-   |compass-short|, see :ref:`download-and-install-compass`.
+   |compass-short|, see :ref:`Download and Install Compass <download-and-install-compass>`.
 
 .. seealso::
 
--- a/source/upgrade.txt
+++ b/source/upgrade.txt
@@ -99,6 +99,14 @@ Procedure
 
 #. Launch the fully-featured edition of |compass|.
 
+.. note::
+
+   If you are running macOS Catalina, or later, you may have to modify
+   your system settings to grant Compass permission to run. For more
+   information on how to grant access permission to Compass, see :ref: `install-compass`.
+
 .. seealso::
 
    :ref:`download-install`
+
+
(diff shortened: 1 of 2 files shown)


source/upgrade.txt
```diff
@@ -99,6 +99,14 @@ Procedure
 
 #. Launch the fully-featured edition of |compass|.
 
+.. note::
+
+   If you are running macOS Catalina, or later versions, you may have to modify
```
jeff-allen-mongo, line 104: [nits] 

- I think we have too many commas here which makes the pace of reading this a little off.

- I don't think we need to say "later versions" we can just say "or later".

Suggest something like:

`If you are running macOS Catalina or later, you may...`

source/upgrade.txt
```diff
@@ -99,6 +99,14 @@ Procedure
 
 #. Launch the fully-featured edition of |compass|.
 
+.. note::
```
jeff-allen-mongo, line 102: So, as is we have this note tucked in the "migration" section of this page. This note doesn't really apply to the migration procedure. Instead, I'd suggest moving this to the bottom of the `Manually Install the Latest Version of Compass` section, since if they are manually installing that's when they are most likely to run into this issue.

jocelyn-mendez:
Hey @jeff-allen-mongo, thanks for the feedback! I have gone through and made the changes, so please feel free to let me know your thoughts when you get a chance.

Updated staging: https://docs-mongodbcom-staging.corp.mongodb.com/compass/jocelynmendez/DOCSP-11049/upgrade.html

jocelyn-mendez:
Hey @mmarcon, I have added the noted regarding permission to run Compass on macOS Catalina or later. Please let me know your thoughts when you get a chance, Thanks!

Updated staging: https://docs-mongodbcom-staging.corp.mongodb.com/compass/jocelynmendez/DOCSP-11049/upgrade.html

Jira: https://jira.mongodb.org/browse/DOCSP-11049

PR #332, opened 2021-03-04
(DOCSP-14404) Add administrator reqs to installing on Windows

JuliaMongo:
[Build](https://workerpool-boxgs.mongodbstitch.com/pages/job.html?jobId=60413cfe75e817aef7af67cb) has no issues
[Jira](https://jira.mongodb.org/browse/DOCSP-14404)
[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/compass/docsworker-xlarge/DOCSP-14404/install/) -- choose the Windows tab.

### DIFF
--- a/source/install.txt
+++ b/source/install.txt
@@ -36,6 +36,9 @@ Select your operating system:
          - 64-bit version of Microsoft Windows 7 or later.
          - MongoDB 3.6 or later.
          - `Microsoft .NET Framework version 4.5 or later <https://www.microsoft.com/en-us/download/details.aspx?id=30653>`__.
+         - Starting the installation as an administrator in your Azure
+           Virtual Desktop Infrastructure (VDI) or on your machine in Microsoft
+           PowerShell.
 
            The |compass-short| installer prompts you to install the
            minimum required version of the .NET framework if it is not
@@ -218,7 +221,10 @@ Select your operating system:
       Install Compass
       ---------------
 
-      1. Double-click the installer file.
+      1. Ensure that you are running PowerShell or your Azure desktop as
+         an administrator.
+         
+      #. Double-click the installer file.
 
       #. Follow the prompts to install |compass-short|. You can
          select the destination of the |compass-short| installation.
--- a/source/install.txt
+++ b/source/install.txt
@@ -36,13 +36,13 @@ Select your operating system:
          - 64-bit version of Microsoft Windows 7 or later.
          - MongoDB 3.6 or later.
          - `Microsoft .NET Framework version 4.5 or later <https://www.microsoft.com/en-us/download/details.aspx?id=30653>`__.
-         - Starting the installation as an administrator in your Azure
-           Virtual Desktop Infrastructure (VDI) or on your machine in Microsoft
-           PowerShell.
-
+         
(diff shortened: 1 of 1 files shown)


feliciahsiehdb:
Looks good

source/install.txt
```diff
@@ -36,11 +36,15 @@ Select your operating system:
          - 64-bit version of Microsoft Windows 7 or later.
          - MongoDB 3.6 or later.
          - `Microsoft .NET Framework version 4.5 or later <https://www.microsoft.com/en-us/download/details.aspx?id=30653>`__.
-
+         
            The |compass-short| installer prompts you to install the
            minimum required version of the .NET framework if it is not
            already installed on your system.
 
+         - Starting the installation as an administrator in your Azure
```
jeff-allen-mongo, line 44: [question]

This makes it sound like I _must_ use Azure VDI or PowerShell to complete the installation, which isn't the case. I was just able to complete the installation on my Windows computer and did not need to use PowerShell.

It sounds like the ticket is suggesting that we update the docs to say that you should only install using Azure VDI or PowerShell, but I worry this may be an unnecessary complication for the majority of users.

I'm wondering if this be reworded to something like:

"Starting the installation as an administrator if you are installing using Azure Virtual Desktop Infrastructure (VDI) or Microsoft PowerShell."

source/install.txt
```diff
@@ -218,7 +222,10 @@ Select your operating system:
       Install Compass
       ---------------
 
-      1. Double-click the installer file.
+      1. Ensure that you are running PowerShell or your Azure desktop as
+         an administrator.
+         
+      #. Double-click the installer file.
```
jeff-allen-mongo, line 229: A similar comment applies here. Should this be more of a note at the beginning of the procedure saying:

"If you are installing using PowerShell or your Azure desktop, ensure that you are running as an administrator."

or similar.

Also, I would think that if we are assuming users are in PowerShell, we would provide a PowerShell command rather than telling them to double click the installer file in step 2? Since double-clicking the installer file would just start the .msi, and not use PowerShell at all.

source/install.txt
JuliaMongo: Makes sense! Will update, Jeff!

source/install.txt
JuliaMongo: Yes, this content is not needed here at all, it only is needed in the prereqs. Will remove from here.

source/install.txt
JuliaMongo: @jeff-allen-mongo please take another look? I made the changes.

jeff-allen-mongo:
LGTM, though I'd appreciate @mmarcon's eyes on this before we merge just to make sure the messaging here is correct. Thank you Julia 🚀

jeff-allen-mongo (approved):

mmarcon:
Uhmm. I am not sure about the way it's phrased. Powershell and Azure VDI are 2 different things - it's like saying "if you are installing using macOS or the terminal". How about something like this:

```
Starting the installation as an administrator if you are doing a silent installation using Microsoft PowerShell or installing on Azure Virtual Desktop Infrastructure (VDI).
```

PR #362, opened 2021-11-16
Mention automatic updates in the list of things isolated does not do

mmarcon:
Also removed crash reporting as we don't collect that info anymore
and made the sentence about data collection more generic.


jeff-allen-mongo:
Hey @mmarcon , thanks for submitting this. I've tweaked this slightly to pull the auto updates into their own row on the table. I merged the change in PR #363

PR #374, opened 2022-02-09
DOCSP-20684 Connect to MongoDB page updated

jocelyn-mendez1:
Updating Connect to MongoDB page to reflect new compass connection form.

JIRA: https://jira.mongodb.org/browse/DOCSP-20684

STAGING: https://docs-mongodbcom-staging.corp.mongodb.com/compass/docsworker-xlarge/DOCSP-20684/connect/


jocelyn-mendez1:
Hi @jeff-allen-mongo, here is the reworked Connect to MongoDB page. Let me know of additional changes that can be made. Thanks!

source/connect.txt
```diff
@@ -49,68 +49,69 @@ Connect
...
    :figwidth: 690px
-   :alt: Connection dialog for pasting your connection string
+   :alt: New connection dialog for pasting your connection string
 
-|compass-short| provides two methods to connect to your deployment,
-either providing your deployment connection string or filling in your
-deployment information in specific fields.
+|compass-short| provides multiple methods to connect to your deployment,
+either providing your deployment connection string or through Advanced 
+Connection Options.
```
jeff-allen-mongo, line 58: I suggest we modify this to make the user the actor in the sentence, rather than Compass. We can try something like:

```
You can connect to your deployment by either:

- Providing your deployment connection string
- Specifying :guilabel:`Advanced Connection Options`
```

source/connect.txt
```diff
@@ -49,68 +49,69 @@ Connect
...
+value is 60000, or 60 seconds. If you are frequently experiencing
+timeouts in |compass-short|, consider setting this option to a
+higher value in your connection string.
 
-      By default, |compass-short|'s default 
-      :manual:`socketTimeoutMS </reference/connection-string/#urioption.socketTimeoutMS>`
-      value is 60000, or 60 seconds. If you are frequently experiencing
-      timeouts in |compass-short|, consider setting this option to a
-      higher value in your connection string.
+For a complete list of the connection string options which
```
jeff-allen-mongo, line 74: I think "which" should be "that" in this case.

source/connect.txt
```diff
@@ -49,68 +49,69 @@ Connect
...
+---------------------------
 
-   .. tab:: Fill in Connection Fields Individually
-      :tabid: individual-fields
+The Advanced Connection Options provide additonal forms of connecting
+|compass-short| deployments.  
 
-      .. include:: /includes/steps/starting-compass-individual-fields.rst
+:doc:`/connect/general-connection`
+  Learn how to select a schema for your deployment.
```
jeff-allen-mongo, line 89: I don't know if "select a schema" is the best choice of language here. "Schema" typically refers to the way you model your data. That's how it's described elsewhere in the docs: https://docs.mongodb.com/manual/data-modeling/

However, I see that "schema" is the language used in the Compass UI. I think we may need to run this by the Compass dev team and ask why "schema" was the word chosen for this option, because I don't think it's the clearest word to use. Something like "Connection String Format" would be more appropriate imo.

For now, I suggest we change this line to something like "Learn how to change your connection string format." I'll start a thread in the #compass-dev Slack channel inquiring about the text used in the Compass UI so we can see if it can / will be updated.

source/connect.txt
jeff-allen-mongo, line 58: After this paragraph, consider adding a quick sentence explaining what the advanced options consist of. For example:

"Advanced connection options allow you to specify authentication, TLS/SSL, and SSH connection options. To learn more, see (link to the Advanced Connection Options section further down on this page)."

jeff-allen-mongo (commented):
@jocelyn-mendez1 , this is looking good! Just a couple suggestions for you.

source/connect.txt
jocelyn-mendez1: Done.

source/connect.txt
jocelyn-mendez1: I updated this to "Connection String Scheme" as was seen in the PR for the change in the #compass-dev Slack.

source/connect.txt
jocelyn-mendez1: Done.

source/connect.txt
```diff
@@ -49,68 +49,77 @@ Connect
...
 
-   .. tab:: Paste Connection String
-      :tabid: paste-connection-string
+- Specifying :guilabel:`Advanced Connection Options`. 
 
-      When you provide a connection string, |compass-short| supports
-      most :manual:`Connection String Options
-      </reference/connection-string/#connection-string-options>`
-      supported by MongoDB.
+Advanced connection options allow you to specify authentication,
```
jeff-allen-mongo, line 62: We should actually probably make this paragraph part of the second bullet.

jeff-allen-mongo (approved):
LGTM % a suggestion.

source/connect.txt
jocelyn-mendez1: Done.

addaleax (approved):

PR #432, opened 2022-10-11
DOCSP-25682): Import favorite connections

jeff-allen-mongo:


jeff-allen-mongo:
Copy review performed on https://github.com/jeff-allen-mongo/docs-compass/pull/1/

PR #443, opened 2022-12-06
(DOCSP-25415): networkTraffic config option

jeff-allen-mongo:
STAGED: https://docs-mongodbcom-staging.corp.mongodb.com/compass/docsworker-xlarge/DOCSP-25415-b/config-file/

BUILD LOG: https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=638f84f9dc742b91f25c2921


jeff-allen-mongo:
@jocelyn-mendez1 that's a good question. I don't have a strong preference here. My reasoning for the structure used in this PR is that having "task" pages for specific options at the same level as a "reference" page for all of the options made sense as an approach, but happy to restructure as part of a future PR if you have other ideas!

PR #445, opened 2022-12-12
DOCSP-27031 Compass CLI options

jmd-mongo:
## STAGING

https://docs-mongodbcom-staging.corp.mongodb.com/compass/docsworker-xlarge/DOCSP-27031/command-line-options/#options

## JIRA

https://jira.mongodb.org/browse/DOCSP-27031

## BUILD LOG

https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=6397abe0dc742b91f2f593bf

### DIFF
--- a/source/command-line-options.txt
+++ b/source/command-line-options.txt
@@ -55,8 +55,9 @@ You can set these options when using the command line:
 
 .. option:: --file <file>
   
-   Specify a file containing a list of connections. These connections
-   are automatically available once |compass| starts.
+   Specify a file containing a :ref:`list of connections 
+   <compass-export-connections-cli>`. These connections are 
+   automatically available once |compass| starts.
 
 .. option:: --help
    
--- a/source/command-line-options.txt
+++ b/source/command-line-options.txt
@@ -39,15 +39,27 @@ You can set these options when using the command line:
 
 .. option:: --autoUpdates
   
-   Enable :ref:`automatic updates <enable-automatic-updates>`.
+   Enable :ref:`automatic updates <enable-automatic-updates>`. Use 
+   ``--no-autoUpdates`` to disable automatic updates. 
+
+   Setting this option overrides the value in the :guilabel:`Settings` 
+   panel.
   
 .. option:: --enableFeedbackPanel
   
-   Enable the feedback panel.
+   Enable the feedback panel. Use ``--no-enableFeedbackPanel`` to 
+   disable automatic updates. 
+
+   Setting this option overrides the value in the :guilabel:`Settings` 
+   panel.
    
 .. option:: --enableMaps
  
-   Enable geographic visualizations.
+   Allow |compass| to make requests to a 3rd party mapping service. Use
+   ``--no-enableMaps`` to disable mapping requests.
+
+   Setting this option overrides the value in the :guilabel:`Settings` 
+   panel.
 
 .. option:: --exportConnections
     
@@ -74,7 +86,7 @@ You can set these options when using the command line:
 .. option:: --networkTraffic
 
    Configure |compass| to only allow outgoing :ref:`network operations 
-   <compass-configure-network-traffic>` to connect to the database.
+   <compass-configure-network-traffic>` to connect to the database. 
 
 .. option:: --passphrase
  
@@ -86,22 +98,35 @@ You can set these options when using the command line:
 
 .. option:: --protectConnectionStrings
 
-   Protect connection string secrets. Passwords in connection strings
-   are displayed as `*****`. Toggle :guilabel:`Edit Connection
-   String` to display the full, unprotected connection string.
+   Hide credentials in connection strings. Passwords in connection 
+   strings are displayed as `*****`. 
+
+   Setting this option overrides the value in the :guilabel:`Settings` 
+   panel.
 
 .. option:: --theme
 
    Specify the |compass| UI theme. The supported themes are ``DARK``,
    ``LIGHT``, and ``OS_THEME``.
+   
+   Setting this option overrides the value in the :guilabel:`Settings` 
+   panel.
 
 .. option:: --trackErrors
 
-   Enable crash reports.
+   Enable sending crash reports. Use ``--no-trackErrors`` to disable
+   sending crash reports.
+   
+   Setting this option overrides the value in the :guilabel:`Settings` 
+   panel.
 
 .. option:: --trackUsageStatistics
 
-   Enable sending usage statistics.
+   Enable sending usage statistics. Use ``--no-trackUsageStatistics``
+   to disable sending usage statistics.
+
+   Setting this option overrides the value in the :guilabel:`Settings` 
+   panel.
 
 .. option:: --version
 
--- a/source/command-line-options.txt
+++ b/source/command-line-options.txt
@@ -95,10 +95,6 @@ overrides the value in the :guilabel:`Settings` panel.
    Allow |compass| to make requests to a 3rd party mapping service. Use
    ``--no-enableMaps`` to disable mapping requests.
 
-.. option:: --lgDarkmode
-
-   Use leafygreen dark mode instead of darkreader.
-
 .. option:: --networkTraffic
 
    Configure |compass| to only allow outgoing :ref:`network operations 
--- a/source/command-line-options.txt
+++ b/source/command-line-options.txt
@@ -86,7 +86,9 @@ You can set these options when using the command line:
 
 .. option:: --protectConnectionStrings
 
-   Protect connection string secrets.
+   Protect connection string secrets. Passwords in connection strings
+   are displayed as `*****`. Toggle :guilabel:`Edit Connection
+   String` to display the full, unprotected connection string.
 
 .. option:: --theme
 
@@ -99,7 +101,7 @@ You can set these options when using the command line:
 
 .. option:: --trackUsageStatistics
 
-   Enable usage statistics.
+   Enable sending usage statistics.
 
 .. option:: --version
 
--- a/source/command-line-options.txt
+++ b/source/command-line-options.txt
@@ -1,5 +1,7 @@
 .. _cli-options:
 
+.. program:: compass
+
 ----------------------------
 Compass Command Line Options
 ----------------------------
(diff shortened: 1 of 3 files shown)


source/command-line-options.txt
```diff
@@ -33,18 +35,77 @@ Use the following syntax to start |compass| and set a configuration option:
 Options
 -------
 
-You can set the following configuration options using the command line:
+You can set these options when using the command line:
```
davemungo, line 38: Are there links available for these topics (like on line 55)?

source/command-line-options.txt
```diff
@@ -33,18 +35,77 @@ Use the following syntax to start |compass| and set a configuration option:
...
+  
+   Enable the feedback panel.
+   
+.. option:: --enableMaps
+ 
+   Enable geographic visualizations.
+
+.. option:: --exportConnections
+    
+   Export :ref:`favorite connections
```
davemungo, line 54: The import / export page is pretty long. Can this link directly to export rather than the top of the page?

source/command-line-options.txt
```diff
@@ -33,18 +35,77 @@ Use the following syntax to start |compass| and set a configuration option:
...
+   Import :ref:`favorite connections 
+   <compass-export-import-cli-landing>`.
+
+.. option:: --lgDarkmode
+
+   Use leafygreen dark mode instead of darkreader.
+
+.. option:: --networkTraffic
+
+   Configure |compass| to not perform outgoing
```
davemungo, line 77: Consider: 
Configure |compass| to only allow outgoing :ref:`network operations <compass-configure-network-traffic>` to connect to the database.

source/command-line-options.txt
```diff
@@ -33,18 +35,77 @@ Use the following syntax to start |compass| and set a configuration option:
...
+
+.. option:: --networkTraffic
+
+   Configure |compass| to not perform outgoing 
+   :ref:`network operations <compass-configure-network-traffic>`
+   other than those to the database.
+
+.. option:: --passphrase
+ 
+   Connection export/import passphrase.
```
davemungo, line 83: Rather than the /, consider something like:

The passphrase that protects the :link:`exported` connection credentials.

source/command-line-options.txt
```diff
@@ -33,18 +35,77 @@ Use the following syntax to start |compass| and set a configuration option:
...
+
+.. option:: --positionalArguments
+
```
davemungo, line 86: This one seems strange. Judging by the description, 'postionalArguments' probably isn't an option flag. These look like positional parameters.

davemungo (commented):
@jmd-mongo thanks for putting the options list together. I left a couple of comments, mostly around additional linking.

source/command-line-options.txt
jmd-mongo: It strikes me as odd as well. It looks like these two might actually be for Chromium/Electron. I've removed these.

```
 ./MongoDB\ Compass\ Beta --positionalArguments 
Type for option "positionalArguments" mismatches: expected array, received boolean (while validating preferences from: Command line)
```

```
--ignoreAdditionalCommandLineFlags (*)     Allow Additional CLI Flags
--positionalArguments                      Specify a Connection String or Connection ID to Automatically Connect
```

source/command-line-options.txt
jmd-mongo: Updated. Thanks!

source/command-line-options.txt
jmd-mongo: Done.

source/command-line-options.txt
jmd-mongo: I've updated these links so that the import-related option points to the Import Connections page and the export-related option points to the Export Connections page.

source/command-line-options.txt
jmd-mongo: I'll take a closer look. Not sure about many of these.

jmd-mongo:
Thanks for the review, @davemungo! I've linked out where applicable. I've also updated the description of the 
`--passphrase` option. Could you give this another look when you have a chance?

source/command-line-options.txt
jmd-mongo: I've linked out where I could find additional info on the Compass docs.

source/command-line-options.txt
```diff
@@ -33,18 +35,74 @@ Use the following syntax to start |compass| and set a configuration option:
...
+ 
+   Enable geographic visualizations.
+
+.. option:: --exportConnections
+    
+   :ref:`Export <compass-export-connections-cli>` favorite connections.
+
+.. option:: --file <file>
+  
+   Specify a file containing a list of connections. These connections
```
davemungo, line 58: We don't have a page for the file itself, perhaps linking to the instructions for creating the file is a good alternative.

Specify a file containing a :ref:`list of connections <compass-export-connections-cli>`. These connections

davemungo (approved):
@jmd-mongo Thanks for the updated links. I left an optional suggestion for one more. 

LGTM

source/command-line-options.txt
jmd-mongo: Good idea!

source/command-line-options.txt
```diff
@@ -33,18 +35,75 @@ Use the following syntax to start |compass| and set a configuration option:
 Options
 -------
 
-You can set the following configuration options using the command line:
+You can set these options when using the command line:
+
+.. option:: --autoUpdates
+  
+   Enable :ref:`automatic updates <enable-automatic-updates>`.
```
addaleax, line 42: I think it’s worth mentioning that:

- Some of these options override values that are otherwise set in the settings (formerly privacy) panel
- For boolean options a `--no` prefix can be used to negate them, e.g. `--no-autoUpdates`

Otherwise `--autoUpdates` is not a particularly useful option on its own, since it’s set by default and can also be disabled from the settings panel directly.

source/command-line-options.txt
```diff
@@ -33,18 +35,75 @@ Use the following syntax to start |compass| and set a configuration option:
...
+.. option:: --trackErrors
+
+   Enable crash reports.
+
+.. option:: --trackUsageStatistics
 
-.. list-table::
-   :header-rows: 1
-   :widths: 15 60
+   Enable usage statistics.
```
addaleax, line 102:
```suggestion
   Enable sending usage statistics.
```

source/command-line-options.txt
```diff
@@ -33,18 +35,75 @@ Use the following syntax to start |compass| and set a configuration option:
...
+ 
+   Specify a passphrase to encrypt the :ref:`exported favorite 
+   connections <compass-export-connections-cli-encrypted>` output file.
+
+   Specify a passphrase to decrypt the :ref:`imported favorite
+   connections <compass-import-connections-cli-encrypted>` file. 
+
+.. option:: --protectConnectionStrings
+
+   Protect connection string secrets.
```
addaleax, line 89: Do you want to include something like the long-form explanation that we include for most options in the settings panel?

source/command-line-options.txt
jmd-mongo: I've updated the options that accept boolean values to document the `--no-` versions as well. 

I've noted that setting options on the command line will override values set in **Settings**.

source/command-line-options.txt
jmd-mongo: I expanded this description with info from the **Settings** panel. Thanks!

source/command-line-options.txt
```diff
@@ -33,18 +35,103 @@ Use the following syntax to start |compass| and set a configuration option:
...
+   connections <compass-export-connections-cli-encrypted>` output file.
+
+   Specify a passphrase to decrypt the :ref:`imported favorite
+   connections <compass-import-connections-cli-encrypted>` file. 
+
+.. option:: --protectConnectionStrings
+
+   Hide credentials in connection strings. Passwords in connection 
+   strings are displayed as `*****`. Toggle :guilabel:`Edit Connection 
+   String` to display the full, unprotected connection string.
```
addaleax, line 103: The point of this option is that you cannot do this anymore if it’s set :)

source/command-line-options.txt
```diff
@@ -33,18 +35,103 @@ Use the following syntax to start |compass| and set a configuration option:
...
+
+   Setting this option overrides the value in the :guilabel:`Settings` 
+   panel.
```
addaleax, line 106: This applies to all command line flags that correspond to settings. Do you want to leave it here only for `--protectConnectionStrings`?

source/command-line-options.txt
jmd-mongo: Got it -- updated!

source/command-line-options.txt
jmd-mongo: I included this bit in descriptions of: 
- `--autoUpdates` 
- `--enableFeedbackPanel`
- `--enableMaps` 
- `--theme`
- `--trackErrors`
- `--trackUsageStatistics`

 Are you asking if we should only mention this once? Thanks!

addaleax (approved):

source/command-line-options.txt
jmd-mongo: Thanks for this note! I've broken the options out into two sub-sections:

- [Command Line Only](https://docs-mongodbcom-staging.corp.mongodb.com/compass/docsworker-xlarge/DOCSP-27031/command-line-options/#command-line-only)
- [Command Line or Configuration File](https://docs-mongodbcom-staging.corp.mongodb.com/compass/docsworker-xlarge/DOCSP-27031/command-line-options/#command-line-and-configuration-file)

source/command-line-options.txt
```diff
@@ -33,18 +35,95 @@ Use the following syntax to start |compass| and set a configuration option:
...
+   disable automatic updates. 
+
+.. option:: --enableMaps
+ 
+   Allow |compass| to make requests to a 3rd party mapping service. Use
+   ``--no-enableMaps`` to disable mapping requests.
+
+.. option:: --lgDarkmode
+
+   Use leafygreen dark mode instead of darkreader.
```
jeff-allen-mongo, line 100: [question]

Will users know what "leafygreen" dark mode is / means? Is there any difference between the two options we might want to highlight here?

jmd-mongo: Thank you for bringing this up. I don't have any other info. I'm going to remove this option.

jeff-allen-mongo: Sounds good, thanks!

PR #446, opened 2022-12-19
DOCSP-27087 adding add pipeline text view section

ianf-mongodb:
DESCRIPTION
--------------

Creating content for 1.35 compass feature "Text mode editor". This includes specifying previous content was using the stage view mode and adding new content detailing how to create an aggregation pipeline in text view mode.

Previous screenshots will be updated in a separate ticket outside of this edit so please maintain commentary to the text view mode feature only.

STAGING
---------

https://docs-mongodbcom-staging.corp.mongodb.com/compass/docsworker-xlarge/DOCSP-27087/aggregation-pipeline-builder/#create-an-aggregation-pipeline---stage-view-mode (small edit for the stage switch, step 1)

https://docs-mongodbcom-staging.corp.mongodb.com/compass/docsworker-xlarge/DOCSP-27087/aggregation-pipeline-builder/#create-an-aggregation-pipeline---text-view-mode

JIRA
-----

https://jira.mongodb.org/browse/DOCSP-27087

BUILD
------

https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=63a0d278dc742b91f2c620f2

PR #447, opened 2022-12-19
DOCSP-27089-Import-Pipeline-Updates

ianf-mongodb:
BUILD
------

https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=63a0da23dc742b91f2c85066

Staging
-------

https://docs-mongodbcom-staging.corp.mongodb.com/compass/docsworker-xlarge/DOCSP-27089/import-pipeline-from-text/

Jira
----

https://jira.mongodb.org/browse/DOCSP-27089


source/import-pipeline-from-text.txt
```diff
@@ -24,18 +27,39 @@ pipeline stages using the Pipeline Builder's :guilabel:`Output` panes.
...
-:manual:`MongoDB query language </tutorial/query-documents/>`
-(i.e. the same syntax used as the ``pipeline`` parameter of the
-:manual:`db.collection.aggregate()
-</reference/method/db.collection.aggregate/>` method). The imported
-pipeline must be an array, even if there is only one stage in the
-pipeline.
+The imported pipeline must be:
+
+- In the :manual:`MongoDB query language </tutorial/query-documents/>`
+  (i.e. the same syntax used as the ``pipeline`` parameter of the
```
jeff-allen-mongo, line 33: Per our style guide recommendations, can we rewrite this sentence to not use "i.e." ?

source/import-pipeline-from-text.txt
```diff
@@ -24,18 +27,39 @@ pipeline stages using the Pipeline Builder's :guilabel:`Output` panes.
...
+  :manual:`db.collection.aggregate()
+  </reference/method/db.collection.aggregate/>` method).
+- Must be an array, even if there is only one stage in the pipeline.
```
jeff-allen-mongo, line 36: We already have "must be" in the lead-in to this bulleted list, so I think we can remove it here.

source/import-pipeline-from-text.txt
```diff
@@ -12,7 +12,10 @@ Import Pipeline from Text
    :depth: 1
    :class: singlecol
 
-*New in version 1.15.0*
+Starting in Compass 1.35, the import pipeline button has been removed
```
jeff-allen-mongo, line 15: I'm not sure this is the best place for this paragraph. As a new user coming in with no context on this procedure, this could potentially read like I'm no longer able to import pipelines, especially if I read too quickly.

Rather than focusing on the negative (what's been removed), maybe this could be reworded a bit to say that the location of the import functionality has moved?

I would also consider making this a note after the intro paragraph, as brand new Compass users don't need to know this information.

source/import-pipeline-from-text.txt
```diff
@@ -24,18 +27,39 @@ pipeline stages using the Pipeline Builder's :guilabel:`Output` panes.
...
+   :style: normal
+
+   .. step:: In the aggregation pipeline pane, click the 
+      :guilabel:`</> TEXT` toggle switch to enable text mode for 
+      pipeline editing.
+
+      .. figure:: /images/compass/agg-builder-textview-toggle-on.png
+         :alt: Toggle Textview Mode on
+
+   .. step:: Type or paste aggregation pipeline syntax text into the
```
jeff-allen-mongo, line 51: Can we simplify this to say: "Type or paste an aggregation pipeline..."

source/import-pipeline-from-text.txt
```diff
@@ -24,18 +27,39 @@ pipeline stages using the Pipeline Builder's :guilabel:`Output` panes.
...
+
+   .. step:: In the aggregation pipeline pane, click the 
+      :guilabel:`</> TEXT` toggle switch to enable text mode for 
+      pipeline editing.
+
+      .. figure:: /images/compass/agg-builder-textview-toggle-on.png
+         :alt: Toggle Textview Mode on
+
+   .. step:: Type or paste aggregation pipeline syntax text into the 
+      text editor. The text editor provides real time linting
```
jeff-allen-mongo, line 52: I believe "real-time" should be hyphenated.

source/import-pipeline-from-text.txt
```diff
@@ -24,18 +27,39 @@ pipeline stages using the Pipeline Builder's :guilabel:`Output` panes.
...
+      .. figure:: /images/compass/agg-builder-textview-toggle-on.png
+         :alt: Toggle Textview Mode on
+
+   .. step:: Type or paste aggregation pipeline syntax text into the 
+      text editor. The text editor provides real time linting 
+      for correct syntax and debugging information.
+
+      .. code-block:: javascript
+
+         [ { "$limit" : 4 } ]
```
jeff-allen-mongo, line 57: This code-block feels slightly out of context as written. Can we provide a lead-in with something like:

```
For example, the following pipeline consists of a single :pipeline:`$limit` stage: 
```

source/import-pipeline-from-text.txt
```diff
@@ -24,18 +27,39 @@ pipeline stages using the Pipeline Builder's :guilabel:`Output` panes.
...
+
+      .. figure:: /images/compass/agg-builder-textview-edit.png
+         :alt: Edit aggregation pipeline
 
-Once you import your pipeline, you can add and modify individual stages
-and see the results reflected in the :guilabel:`Output` of each respective stage.
+Once you import your pipeline, you can click the :guilabel:`{} STAGES`
```
jeff-allen-mongo, line 62: [nit]

I know this copy was preexisting, but we should replace "once" with "after".

source/import-pipeline-from-text.txt
```diff
@@ -24,18 +27,39 @@ pipeline stages using the Pipeline Builder's :guilabel:`Output` panes.
...
+
+         [ { "$limit" : 4 } ]
+
+      .. figure:: /images/compass/agg-builder-textview-edit.png
+         :alt: Edit aggregation pipeline
 
-Once you import your pipeline, you can add and modify individual stages
-and see the results reflected in the :guilabel:`Output` of each respective stage.
+Once you import your pipeline, you can click the :guilabel:`{} STAGES` 
+switch to return to stage view mode and modify individual stages to
```
jeff-allen-mongo, line 63: This sentence is a little long. Maybe we can break it up:

"... click the stages switch to return to stage view mode. In this mode, you can modify..."

ianf-mongodb:
Thanks for the feedback @jeff-allen-mongo, this is ready for re-review.

source/import-pipeline-from-text.txt
```diff
@@ -12,30 +12,58 @@ Import Pipeline from Text
...
-*New in version 1.15.0*
-
 You can import aggregation pipelines from plain text into
 the :ref:`Aggregation Pipeline Builder <compass-agg-builder>` to
 easily modify and execute your pipelines. Importing a plain text
 aggregation pipeline shows how each stage of the pipeline
 affects the output, and illustrates the effects of modifying specific
 pipeline stages using the Pipeline Builder's :guilabel:`Output` panes.
 
+Starting in Compass 1.35, the import pipeline button has been removed
```
jeff-allen-mongo, line 22: Perhaps a `.. note::` admonition would be good here, as this kind of feels like an aside to me.

source/import-pipeline-from-text.txt
```diff
@@ -12,30 +12,58 @@ Import Pipeline from Text
...
-:manual:`MongoDB query language </tutorial/query-documents/>`
-(i.e. the same syntax used as the ``pipeline`` parameter of the
-:manual:`db.collection.aggregate()
-</reference/method/db.collection.aggregate/>` method). The imported
-pipeline must be an array, even if there is only one stage in the
-pipeline.
+The imported pipeline must be:
+
+- In the :manual:`MongoDB query language </tutorial/query-documents/>`,
+  this is the same syntax used as the ``pipeline`` parameter of the
```
jeff-allen-mongo, line 34: [nit]

```suggestion
  this is the same syntax used in the ``pipeline`` parameter of the
```

source/import-pipeline-from-text.txt
```diff
@@ -12,30 +12,58 @@ Import Pipeline from Text
...
+  :manual:`db.collection.aggregate()
+  </reference/method/db.collection.aggregate/>` method.
```
jeff-allen-mongo, line 36: [question]

Does `:method:` work here in place of the `:manual:` directive?

```
:method:`aggregate() <db.collection.aggregate()>`
```

source/import-pipeline-from-text.txt
```diff
@@ -12,30 +12,58 @@ Import Pipeline from Text
...
 
 Procedure
 ---------
 
-.. include:: /includes/steps/import-pipeline-from-text.rst
+.. procedure::
+   :style: normal
+
+   .. step:: In the aggregation pipeline pane, click the 
+      :guilabel:`</> TEXT` toggle switch to enable text mode for
```
jeff-allen-mongo, line 46: I left a similar comment on PR #446 but missed it here - can we use initial case `Text` here rather than all-caps? Same comment applies to other all-caps UI elements in this update.

jeff-allen-mongo (approved):
@ianf-mongodb LGTM with a few final comments.

mabaasit (approved):

PR #456, opened 2023-01-25
DOCSP-26611 Add maxtimeMS

ianf-mongodb:
## DESCRIPTION

Adding `maxTimems` to the aggregation page and additional screenshots. Possible that the general "Documents" maxtimeMS also needs to be documented, currently it is not perhaps a small page after  [query/limit](https://docs-mongodbcom-staging.corp.mongodb.com/compass/docsworker-xlarge/DOCSP-26611/query/limit/) "Limit Execution Time of Queries".

## STAGING

https://docs-mongodbcom-staging.corp.mongodb.com/compass/docsworker-xlarge/DOCSP-26611/aggregation-pipeline-builder/#set-maxtimems-for-aggregations-queries

## JIRA 

https://jira.mongodb.org/browse/DOCSP-26611

## BUILD

https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=63d17699817860b93f1a42ad


source/aggregation-pipeline-builder.txt
```diff
@@ -183,6 +183,39 @@ Open a Saved Pipeline
    To learn more, see 
    :ref:`viewing saved aggregations<saved-query-view>` for details.
 
+.. _set-maxtime-ms-agg:
+
+Set ``maxTimeMS`` for Aggregations Queries
+------------------------------------------
+
+Use the ``maxTimeMS`` parameter on the aggregation tab to put an
```
ajhuh-mdb, line 191: [suggestion/nit] suggesting a stronger verb such as "specify" for setting the parameter value

source/aggregation-pipeline-builder.txt
```diff
@@ -183,6 +183,39 @@ Open a Saved Pipeline
...
+
+   .. step:: In the aggregation pipeline pane, click the 
+      :guilabel:`More Options`.
+
+      .. figure:: /images/compass/agg-builder-click-more-options.png
+         :alt: Toggle more options
+
+   .. step:: Enter a value next to :guilabel:`Max Time MS` to set a 
+      maximum amount of time in milliseconds that an aggregation 
+      pipeline can run. For a ``5`` second limit enter ``5000``.
```
ajhuh-mdb, line 208: [fix] missing comma
```suggestion
      pipeline can run. For a ``5`` second limit, enter ``5000``.
```

source/aggregation-pipeline-builder.txt
```diff
@@ -183,6 +183,39 @@ Open a Saved Pipeline
...
+
+      .. figure:: /images/compass/agg-builder-enter-maxtime-ms.png
+         :alt: Enter a maximum timeout value 
+
+   .. step:: Any aggregation pipeline that goes over the time limit
```
ajhuh-mdb, line 213: [suggestion/nit] It seems like "operation" might be better suited here since the timeout is dependent on the operation that runs, rather than the pipeline itself?
```suggestion
   .. step:: Any aggregation operation that goes over the time limit 
```

ajhuh-mdb, line 213: [question/suggestion] I feel like this statement about the error message would be better suited outside the procedure directive / not as a step since it's not particularly a step you have to take to set the `maxTimeMS` option. 

Can take it out of the step and instead say something along the lines of: 
"If your aggregation operation goes over the time limit, |compass-short| raises the following error: "

ajhuh-mdb (commented):
Thanks for adding this section @ianf-mongodb! The screenshots here were especially helpful since the `More Options` panel could be easily missed without any indicator. Just have a few small wording/format suggestions.

source/aggregation-pipeline-builder.txt
ianf-mongodb: Its funny, I originally had this as operation and think I changed it right at the end. Updated!

ianf-mongodb: I think you are correct I have staged it both ways, its not part of the procedure and more of a demonstration piece of information. I have removed this step and updated.

ianf-mongodb (commented):
@ajhuh-mdb Thanks for the suggestions Alison, I have updated and this is ready for re-review.

source/aggregation-pipeline-builder.txt
ajhuh-mdb: Haha that's definitely happened to me before, too. Terminology always gets mixed up in my head 😵‍💫

ajhuh-mdb (approved):
LGTM! 

Thanks for making these edits — these look great 😄 
Also, thanks for adding the :guilabel: directive for the **Aggregations** tab. I totally missed that the first round 😅

ianf-mongodb:
Hi @addaleax, could you please have a look at this edit when you get a moment. Thanks!

addaleax (approved):
The additions here look good! For clarity though, DOCSP-26611 is about the additional item in the Compass settings, not the field for the individual query/aggregation tabs. It’s also okay to not document that if it doesn’t seem worth documenting

ianf-mongodb:
> The additions here look good! For clarity though, DOCSP-26611 is about the additional item in the Compass settings, not the field for the individual query/aggregation tabs. It’s also okay to not document that if it doesn’t seem worth documenting

Thanks Anna, I will add the additional setting in a new ticket with scope of documenting the settings menu and adding a TOC entry.

PR #460, opened 2023-02-14
DOCSP-26519 Enable Shell UI Settings

ajhuh-mdb:
## DESCRIPTION
Add information about disabling/enabling the embedded shell on the settings interface. Opted to focus on disabling since there are other relevant circumstances to note alongside it (i.e. limitations of set read-only mode)

## JIRA 
https://jira.mongodb.org/browse/DOCSP-26519

## STAGING 
https://docs-mongodbcom-staging.corp.mongodb.com/compass/docsworker-xlarge/DOCSP-26519-enable-shell-ui-settings/embedded-shell/#disable-the-embedded-mongodb-shell

## BUILD 
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=63ecf46fdcefe97797aa76b3

### DIFF
--- a/source/embedded-shell.txt
+++ b/source/embedded-shell.txt
@@ -84,8 +84,7 @@ To disable the embedded MongoDB shell:
       |compass-short| opens a dialog box where configure your |compass| 
       settings.
 
-   .. step:: Toggle :guilabel:`Enable MongoDB Shell` to disable the embedded 
-      shell on |compass-short|.
+   .. step:: Toggle :guilabel:`Enable MongoDB Shell`.
 
       If you select :guilabel:`Set Read-Only Mode`, |compass-short| 
       automatically disables the :guilabel:`Enable MongoDB Shell` setting.
--- a/source/embedded-shell.txt
+++ b/source/embedded-shell.txt
@@ -85,12 +85,12 @@ To disable the embedded MongoDB shell:
       settings.
 
    .. step:: Toggle :guilabel:`Enable MongoDB Shell` to disable
-      :binary:`mongosh` on |compass-short|
+      :binary:`mongosh` on |compass-short|.
 
       If you select :guilabel:`Set Read-Only Mode`, |compass-short| 
       automatically disables the :guilabel:`Enable MongoDB Shell` setting.
 
-   .. step:: Click :guilable:`Save`. 
+   .. step:: Click :guilabel:`Save`. 
 
 Learn More
 ~~~~~~~~~~
--- a/source/embedded-shell.txt
+++ b/source/embedded-shell.txt
@@ -66,6 +66,32 @@ an :ref:`aggregation pipeline <aggregation-pipeline>`:
         { $match: { "last_name": "Smith" } } // press Shift + Enter
       ] )                                    // Press Enter
 
+Disable the Embedded MongoDB Shell
+----------------------------------
+
+You can disable the embedded MongoDB shell in |compass-short| to restrict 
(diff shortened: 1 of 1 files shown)


source/embedded-shell.txt
```diff
@@ -66,6 +66,31 @@ an :ref:`aggregation pipeline <aggregation-pipeline>`:
...
+To disable the embedded MongoDB shell: 
+
+.. procedure::
+   :style: normal
+
+   .. step:: In the |compass-short| top menu bar, click :guilabel:`Help`.
+
+   .. step:: In the :guilabel:`Help` menu, click :guilabel:`Settings`.
+      
+      |compass-short| opens a dialog box where configure your |compass|
```
ianf-mongodb, line 84: *Typo*

"where configure" -> where you configure

source/embedded-shell.txt
```diff
@@ -66,6 +66,31 @@ an :ref:`aggregation pipeline <aggregation-pipeline>`:
...
+
+   .. step:: In the :guilabel:`Help` menu, click :guilabel:`Settings`.
+      
+      |compass-short| opens a dialog box where configure your |compass| 
+      settings.
+
+   .. step:: Toggle :guilabel:`Enable MongoDB Shell`.
+
+      If you select :guilabel:`Set Read-Only Mode`, |compass-short| 
+      automatically disables the :guilabel:`Enable MongoDB Shell` setting.
```
ianf-mongodb, line 90: **Suggestion**

Update disables -> unchecks, to avoid the paradox contradiction on line 90.

source/embedded-shell.txt
```diff
@@ -66,6 +66,31 @@ an :ref:`aggregation pipeline <aggregation-pipeline>`:
         { $match: { "last_name": "Smith" } } // press Shift + Enter
       ] )                                    // Press Enter
 
+Disable the Embedded MongoDB Shell
+----------------------------------
+
+You can disable the embedded MongoDB shell in |compass-short| to restrict
```
ianf-mongodb, line 72: Im a little confused by " to restrict users from running unauthorized commands on :binary:`mongosh`."

This setting is the client specific one that affects only the current users Compass client.
If a MongoDB admin were to facilitate the install of Compass with a global config file this could be true, but its not obvious who the "you" is in this context.

I would argue its also unclear in our current config page: https://www.mongodb.com/docs/compass/current/config-file/. 

Perhaps this could be shortened to:

"You can disable the embedded MongoDB shell in |compass-short|"

ianf-mongodb (commented):
@ajhuh-mdb Hi Ali, nice minimalist write-up, I left a few suggestions.

source/embedded-shell.txt
ajhuh-mdb: Good catch. Updated!

source/embedded-shell.txt
ajhuh-mdb: Ah that's a great point. Updated.

ajhuh-mdb:
Hi @ianf-mongodb 👋 
Thanks for the close look on this PR! Just finished making the updates/changes you suggested, so this should be ready for another look. 

Updated build: 
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=63ed30fcdcefe97797b9d574

ianf-mongodb (approved):
Thanks for the updates @ajhuh-mdb, this looks great, nice job! LGTM.

alenakhineika (approved):

PR #497, opened 2023-05-23
DOCSP-29483 Stage Wizard Aggregation Stages

ajhuh-mdb:
## DESCRIPTION
Listing out stages available for use with the Stage Wizard.

QUESTION FOR REVIEWER: 
- Would it be better to put this information in the table with more in-depth descriptions of each use case as [outlined in the PD](https://docs.google.com/document/d/1yeGBHgb-nupd5PtA_9DuHX14yIoTngObKKZoTgLKPWM/edit#heading=h.ea8wetjwaqli) or would that create too much clutter considering those explanations are available in the UI text?

## STAGING
https://docs-mongodbcom-staging.corp.mongodb.com/compass/docsworker-xlarge/DOCSP-29483-stage-wizard-use-cases/aggregation-pipeline-builder/#select-an-aggregation-pipeline-stage-use-case.

## JIRA
https://jira.mongodb.org/browse/DOCSP-29483

## BUILD LOG
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=646ccf05df0f8173e9f5da99

## Self-Review Checklist

### DIFF
--- a/source/aggregation-pipeline-builder.txt
+++ b/source/aggregation-pipeline-builder.txt
@@ -202,10 +202,11 @@ your chosen view mode:
 
          .. step:: Select an aggregation pipeline stage use case.
 
-            On the :guilabel:`Stage Wizard` card, click or drag the stage use 
-            case for the first stage of the pipeline. |compass-short| populates 
-            the stage card with a form that corresponds with the selected 
-            aggregation pipeline stage.
+            On the :guilabel:`Stage Wizard` card, select a stage use 
+            case for the first stage of the pipeline. You can click the card to 
+            add it to the end of your pipeline or drag it to your preferred 
+            position. |compass-short| populates the stage card with a form that 
+            corresponds with the selected aggregation pipeline stage.
 
             The Stage Wizard use cases include the following aggregation stages:
 
--- a/source/aggregation-pipeline-builder.txt
+++ b/source/aggregation-pipeline-builder.txt
@@ -207,6 +207,18 @@ your chosen view mode:
             stage card with a form that corresponds with the selected 
             aggregation pipeline stage.
 
+            You can use the following aggregation stages in the Stage Wizard:
+
+            - :pipeline:`$sort`
+
+            - :pipeline:`$project`
+
+            - :pipeline:`$lookup`
+
+            - :pipeline:`$match`
+
+            - :pipeline:`$group`
+
          .. step:: Fill in your pipeline stage. 
 
             Fill in the form for your selected stage and click 
--- a/source/aggregation-pipeline-builder.txt
+++ b/source/aggregation-pipeline-builder.txt
@@ -202,9 +202,9 @@ your chosen view mode:
 
          .. step:: Select an aggregation pipeline stage use case.
 
-            On the :guilabel:`Stage Wizard` card, select the stage use case for 
-            the first stage of the pipeline. |compass-short| populates the the 
-            stage card with a form that corresponds with the selected 
+            On the :guilabel:`Stage Wizard` card, click or drag the stage use 
+            case for the first stage of the pipeline. |compass-short| populates 
+            the stage card with a form that corresponds with the selected 
             aggregation pipeline stage.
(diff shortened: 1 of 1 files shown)


source/aggregation-pipeline-builder.txt
```diff
@@ -207,6 +207,18 @@ your chosen view mode:
             stage card with a form that corresponds with the selected
```
mdb-ashley, line 207: [suggestion] Just a note that "New in version 1.14.0" appears to be missing a period.

source/aggregation-pipeline-builder.txt
```diff
@@ -207,6 +207,18 @@ your chosen view mode:
             stage card with a form that corresponds with the selected 
             aggregation pipeline stage.
 
+            You can use the following aggregation stages in the Stage Wizard:
```
mdb-ashley, line 210: [discussion] To your point, I suspect that adding much more information here would crowd the page. If the info is available in the UI, I think that's fine. If we really need this info in the docs, I'd suggest it might be time to break each of these ways to build an Aggregation Pipeline out into its own page.

mdb-ashley (approved):
LGTM! Thanks for your work on this. I'll be interested in the external reviewer's opinion as well.

source/aggregation-pipeline-builder.txt
ajhuh-mdb: Just to double check, are you referring to it being missing after the last `0`?

mdb-ashley: Yes!

source/aggregation-pipeline-builder.txt
```diff
@@ -207,6 +207,18 @@ your chosen view mode:
...
+
+            - :pipeline:`$group`
```
mabaasit, line 212: By mentioning these here, users may expect that they can create a complex stage for these stage operators - which is not the case. Stage wizard use-cases offer (at least for now) a basic implementation of these stage operators.

source/aggregation-pipeline-builder.txt
ajhuh-mdb: Noted. We'll see what Julia thinks about this.

source/aggregation-pipeline-builder.txt
mabaasit: that looks better. thanks :)

source/aggregation-pipeline-builder.txt
ajhuh-mdb: @mdb-ashley Julia said this looks good as is!

ajhuh-mdb:
Hey @mabaasit 👋 
Just checked with Julia, and she says that the listing of the stages that are used within the Stage Wizard use cases should suffice. Made some edits to the wording in the intro text in line 97 to emphasize the note that I previously added then deleted per her request, as well as the intro text for the list of stages. LMK what you think and thanks!

source/aggregation-pipeline-builder.txt
```diff
@@ -228,11 +240,9 @@ your chosen view mode:
             To add more aggregation stages to your pipeline, repeat steps 3 
             and 4 for each additional stage. 
             
-            .. note:: 
-               
-               By default, the Stage Wizard adds each new stage to the end of 
-               your pipeline. You can change the order of pipeline stages by 
-               dragging the header of each stage card. 
+            By default, the Stage Wizard adds each new stage to the end of
```
mabaasit, line 243: `By default` here refers to clicking a card in the sidebar. However, users can also drag a card and drop it anywhere in between stage and in that case its not the last one.

ajhuh-mdb: Ah noted — will clarify this ability across both.

ajhuh-mdb: Adding this to behavior update under step 3, telling users that they can click a card to add it to the end or drag it to their preferred position.

PR #529, opened 2023-08-02
DOCSP-30599 Export Aggregation Results in Compass

ajhuh-mdb:
## DESCRIPTION
Create an "Export Aggregation Pipeline Results" task page. Source material from [here](https://www.mongodb.com/docs/compass/current/aggregation-pipeline-builder/#export-aggregation-results).

## STAGING
https://docs-mongodbcom-staging.corp.mongodb.com/compass/docsworker-xlarge/DOCSP-30599-export-aggregation-results/agg-pipeline-builder/export-pipeline-results/

## JIRA
https://jira.mongodb.org/browse/DOCSP-30599

## BUILD LOG
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=64cabc9187b3b360465de08e

## Self-Review Checklist


source/agg-pipeline-builder/export-pipeline-results.txt
```diff
@@ -0,0 +1,64 @@
...
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+To use your aggregation data in other external tools, you can export your
```
ajhuh-mdb, line 15: Note: I'm not completely sure if this would be the primary use case for exporting, but it's the one that came to mind for me. Thoughts?

source/agg-pipeline-builder/export-pipeline-results.txt
```diff
@@ -0,0 +1,64 @@
...
+      - Relaxed Extended JSON 
+
+      - Canonical Extended JSON 
+
+      To learn more about JSON formats, see :ref:`export-data-collection`.
+
+   .. step:: Click Export
+
+      In the pop-up modal, enter a name for your export file, specify the file 
+      destination, and click :guilabel:`Select` to export your pipeline results.
```
jeff-allen-mongo, line 57: [suggestion]

Going through this procedure (admittedly quickly), I didn't realize that we were talking about specifying the destination in the _second_ modal that appears, and I got a little confused.

I wonder if it might help to have "Click Export" be a step with no context, then add another step called "Specify export destination" (or similar), and include this context there. It might also help to say something like "In the new popup modal" or "In the second popup modal" to make sure users are in the right place.

source/agg-pipeline-builder/export-pipeline-results.txt
jeff-allen-mongo: I think this makes sense as a use case. I just might tweak "aggregation data" to "aggregated data".

One other possible use case to mention might be to export pipeline results and then import those results to a different collection. (Note: I'm not sure how useful this is, given that users can accomplish the same thing with the $out stage.)

jeff-allen-mongo (changes requested):
@ajhuh-mdb thanks for these updates! I left some suggestions, mostly on how we might be able to improve the preexisting copy a bit.

source/agg-pipeline-builder/export-pipeline-results.txt
ajhuh-mdb: Gotcha. In that case, I think I'll stick to this.

source/agg-pipeline-builder/export-pipeline-results.txt
```diff
@@ -0,0 +1,64 @@
...
+   .. step:: Click the Export button
+
+      In the top right of the aggregation pipeline builder, click the 
+      :guilabel:`Export` button to open the modal.
+      
+      .. figure:: /images/compass/agg-export.png
+         :alt: Export button on aggregation pipeline
+         :width: 696px
+
+   .. step:: Choose the appropriate file type
```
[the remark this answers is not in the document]
ajhuh-mdb: Updated.

source/agg-pipeline-builder/export-pipeline-results.txt
```diff
@@ -0,0 +1,64 @@
...
+      the :guilabel:`Advanced JSON Format` drop-down menu and select from the 
+      following extended JSON formats: 
+
+      - Default Extended JSON 
+
+      - Relaxed Extended JSON 
+
+      - Canonical Extended JSON 
+
+      To learn more about JSON formats, see :ref:`export-data-collection`.
```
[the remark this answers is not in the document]
ajhuh-mdb: That's a good suggestion! Wasn't sure if it'd be redundant to include across both pages, but it definitely would be easier to just have the table there. Adding!

source/agg-pipeline-builder/export-pipeline-results.txt
ajhuh-mdb: Ah noted!

ajhuh-mdb: That makes sense given they are pretty distinct steps from each other in the actual user experience.

source/agg-pipeline-builder/export-pipeline-results.txt
```diff
@@ -0,0 +1,64 @@
...
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+To use your aggregation data in other external tools, you can export your 
+pipeline results from |compass| as a JSON or CSV file. You can export your 
+pipeline results at any point while creating or editing your pipeline.
```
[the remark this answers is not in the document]
ajhuh-mdb: The main point I was trying to drive was that you don't need to have a completed pipeline to be able to export results, but on second look I see how that sentence as is looks a bit confusing. Opting to remove it since it doesn't seem crucial to the user experience to know this.

ajhuh-mdb:
Thanks for the first round of feedback @jeff-allen-mongo ! This should be ready for another look.

source/agg-pipeline-builder/export-pipeline-results.txt
```diff
@@ -0,0 +1,65 @@
...
+   :class: singlecol
+
+To use your aggregated data in other external tools, you can export your 
+pipeline results from |compass| as a JSON or CSV file. 
+
+About This Task 
+---------------
+
+Avoid exporting to CSV files when possible. CSV files may lose type information 
+and are not suitable for backing up your data.
```
jeff-allen-mongo, line 22: [suggestion]

I think rather than a blanket warning of basically "avoid CSVs all the time", it might be helpful to narrow in on their downsides. Even though you shouldn't use CSVs for backups, I think there are other use cases that we might be discounting with this copy.

Maybe we can tweak this to be something like: "Avoid using CSV exports to back up your data. CSV files may lose type information and are not suitable for backups and restores."

jeff-allen-mongo (approved):
LGTM! I just have a small suggestion about the About this Task section. Good to send off to external review!

source/agg-pipeline-builder/export-pipeline-results.txt
ajhuh-mdb: Ah gotcha! Thanks for the suggestion. Updated!

source/agg-pipeline-builder/export-pipeline-results.txt
```diff
@@ -0,0 +1,65 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+To use your aggregated data in other external tools, you can export your 
+pipeline results from |compass| as a JSON or CSV file. 
+
+About This Task
```
sarah-olson-mongodb, line 18: (nit)
```suggestion
About this Task 
```

sarah-olson-mongodb (approved):

PR #531, opened 2023-08-14
DOCSP-32094 Add more query examples

jeff-allen-mongo:
## DESCRIPTION

Added two new examples to the Compass query bar page:

- Match on multiple conditions ($and)
- Match with $or.

As part of these updates, I reformatted the query bar examples to not use tabs. This way, the example headings appear in the right-hand toc and act as jump-list so users can quickly find examples. The new examples are noted in the in-line comments.

## STAGING

https://docs-mongodbcom-staging.corp.mongodb.com/compass/docsworker-xlarge/DOCSP-32094/query/filter/#examples

## JIRA

https://jira.mongodb.org/browse/DOCSP-32094

## BUILD LOG

https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=64da4b5b67dbe910108ee119

## Self-Review Checklist

- [x] Is this free of any warnings or errors in the RST?
- [x] Is this free of spelling errors?
- [x] Is this free of grammatical errors?
- [x] Is this free of staging / rendering issues?
- [x] Are all the links working?

### DIFF
--- a/source/query/filter.txt
+++ b/source/query/filter.txt
@@ -101,7 +101,7 @@ sample data into your MongoDB deployment, perform the following steps:
    cluster with sample data. The following example queries filter the
    sample JSON documents provided on this page, and do not work with 
 
-Match Documents on a Single Condition
+Match Documents by a Single Condition
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
          
 The following :ref:`query filter <query-bar-filter>` finds
@@ -125,7 +125,7 @@ The query returns the following document:
       "dateCreated": { "$date": "2003-03-26" }
    }
 
-Match Documents on Multiple Conditions
+Match Documents by Multiple Conditions
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The following query filter finds all documents where ``scores`` array
--- a/source/query/filter.txt
+++ b/source/query/filter.txt
@@ -53,45 +53,52 @@ Examples
 The examples on this page use a small example dataset. To import the
 sample data into your MongoDB deployment, perform the following steps:
 
-1. Copy the following documents to your clipboard:
+.. procedure::
+   :style: normal
 
-   .. code-block:: JSON
+   .. step:: Copy the sample documents
+      
+      Copy the following documents to your clipboard:
 
-      [
-         {
-            "name": "Andrea Le",
-            "email": "andrea_le@fake-mail.com",
-            "version": 5,
-            "scores": [ 85, 95, 75 ],
-            "dateCreated": { "$date": "2003-03-26" }
-         },
-         {
-            "email": "no_name@fake-mail.com",
-            "version": 4,
-            "scores": [ 90, 90, 70 ],
-            "dateCreated": { "$date": "2001-04-15" }
-         },
-         {
-            "name": "Greg Powell",
-            "email": "greg_powell@fake-mail.com",
-            "version": 1,
-            "scores": [ 65, 75, 80 ],
-            "dateCreated": { "$date": "1999-02-10" }
-         }
-      ]
+      .. code-block:: JSON
 
-#. In |compass-short|, use the left navigation panel to select the 
-   database and the collection you want to import the data to.
+         [
+            {
+               "name": "Andrea Le",
+               "email": "andrea_le@fake-mail.com",
+               "version": 5,
+               "scores": [ 85, 95, 75 ],
+               "dateCreated": { "$date": "2003-03-26" }
+            },
+            {
+               "email": "no_name@fake-mail.com",
+               "version": 4,
+               "scores": [ 90, 90, 70 ],
+               "dateCreated": { "$date": "2001-04-15" }
+            },
+            {
+               "name": "Greg Powell",
+               "email": "greg_powell@fake-mail.com",
+               "version": 1,
+               "scores": [ 65, 75, 80 ],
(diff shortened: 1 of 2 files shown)


source/query/filter.txt
```diff
@@ -47,6 +47,251 @@ Set Query Filter
...
+   {
+      "_id": { "$oid": "5e349915cebae490877d561d" },
+      "name": "Andrea Le",
+      "email": "andrea_le@fake-mail.com",
+      "version": 5,
+      "scores": [ 85, 95, 75 ],
+      "dateCreated": { "$date": "2003-03-26" }
+   }
+
+Match Documents by Multiple Conditions
```
jeff-allen-mongo, line 128: This is the newly-added example.

source/query/filter.txt
```diff
@@ -47,6 +47,284 @@ Set Query Filter
...
+   {
+      "_id": { "$oid": "5e349915cebae490877d561d" },
+      "name": "Andrea Le",
+      "email": "andrea_le@fake-mail.com",
+      "version": 5,
+      "scores": [ 85, 95, 75 ],
+      "dateCreated": { "$date": "2003-03-26" }
+   }
+
+Match Documents by Multiple Conditions ($and)
```
jeff-allen-mongo, line 128: New example.

source/query/filter.txt
```diff
@@ -47,6 +47,284 @@ Set Query Filter
...
+   {
+      "_id": { "$oid":"5a9427648b0beebeb69579cf" },
+      "name": "Greg Powell",
+      "email": "greg_powell@fake-mail.com",
+      "version": 1,
+      "scores": [ 65, 75, 80 ],
+      "dateCreated": { "$date": "1999-02-10" }
+   }
+
+Match Documents by Multiple Possible Conditions ($or)
```
jeff-allen-mongo, line 152: New example.

source/query/filter.txt
```diff
@@ -47,6 +47,283 @@ Set Query Filter
...
+
+.. code-block:: JSON
+   :copyable: false
+
+   [
+      {
+         "_id": { "$oid":"5e349915cebae490877d561e" },
+         "email": "no_name@fake-mail.com",
+         "version": 4,
+         "scores": [ 90, 90, 75 ],
```
jmd-mongo, line 206: [typo]

The value for this field should be [ 90, 90, 70 ] according to the inserted documents.

source/query/filter.txt
```diff
@@ -47,6 +47,283 @@ Set Query Filter
...
```
jmd-mongo, line 245: Similar comment as on line 206.

source/query/filter.txt
```diff
@@ -47,6 +47,283 @@ Set Query Filter
...
+         "email": "andrea_le@fake-mail.com",
+         "version": 5,
+         "scores": [ 85, 95, 75 ],
+         "dateCreated": { "$date": "2003-03-26" }
+      },
+      {
+         "_id": { "$oid": "5e349915cebae490877d561e" },
+         "email": "no_name@fake-mail.com",
+         "version": 4,
+         "scores": [ 90, 90, 75 ],
```
jmd-mongo, line 292: Similar comment as on line 206.

source/query/filter.txt
```diff
@@ -47,6 +47,283 @@ Set Query Filter
 
    .. include:: /includes/extracts/query-bar-results.rst
 
+Examples
```
jmd-mongo, line 50: [praise]

I think this much more scannable than the previous version and the right-hand nav titles are a clear win. 👍

source/query/filter.txt
```diff
@@ -47,6 +47,283 @@ Set Query Filter
...
+--------
+
+The examples on this page use a small example dataset. To import the
+sample data into your MongoDB deployment, perform the following steps:
+
+1. Copy the following documents to your clipboard:
```
jmd-mongo, line 56: [question]

I'm curious if there's any benefit to using a `procedure::` directive as opposed to using an ordered list here. 

Either way seems reasonable, I just wanted to note it.

jmd-mongo (commented):
Thanks for the improvements and updates here, @jeff-allen-mongo! 

I noted some small issues and have also left a question for your consideration.

Thanks!
Joe

source/query/filter.txt
jeff-allen-mongo: Good idea. I'll update to use the procedure directive.

source/query/filter.txt
jeff-allen-mongo: Fixed.

source/query/filter.txt
jeff-allen-mongo: Fixed.

source/query/filter.txt
jeff-allen-mongo: Fixed.

source/query/filter.txt
jeff-allen-mongo: Ah, so using the procedure directive caused the step titles to appear in the right-hand toc, which results in a confusing / messy presentation.

I think if the steps didn't appear in the toc, the procedure directive would improve the presentation of the main page, but I don't think it works well in this case. Thanks for the suggestion!

jmd-mongo: Got it -- thanks for the info!

jmd-mongo (approved):
Thanks for the updates and giving the `procedure::` directive a shot. This one LGTM!

Thanks,
Joe

PR #542, opened 2023-09-06
DOCSP-32564 Fix Redirects + Missing Screenshots

ajhuh-mdb:
## DESCRIPTION
Screenshots were reported as missing bc the page doesn't exist anymore and didn't have a proper redirect.

## STAGING
N/A for redirects

## JIRA
https://jira.mongodb.org/browse/DOCSP-32564

## BUILD LOG
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=64f8858924fcc731b4c36fed

## Self-Review Checklist


config/redirects
```diff
@@ -8,6 +8,8 @@ raw: docs/compass/ -> ${base}/current/
 raw: docs/compass/manual/aggregation-pipeline-builder -> ${base}/current/aggregation-pipeline-builder/
 raw: docs/compass/current/query/favorite/ -> ${base}/current/query/queries/
 raw: docs/compass/current/dark-mode/ -> ${base}/current/settings/settings-reference/#settings
+raw: docs/compass/master/dark-mode/ -> ${base}/current/settings/settings-reference/#settings
 raw: docs/compass/current/config-file/config-file-options/ -> ${base}/current/settings/config-file/
+raw: docs/compass/master/config-file/config-file-options/ -> ${base}/current/settings/config-file/
```
jeff-allen-mongo, line 13: [out of scope]

These updates look good. I think we may want to run this file by the DOP folks at some point to see how we need to refactor this to avoid needing to list each redirect twice (once for `master` and once for `current`). I'm assuming it's something to do with symlinks, but ideally we'd be able to catch all versions with `[*]`. I'm not sure if there's something preventing us from doing that now.

jeff-allen-mongo (approved):
LGTM!

config/redirects
ajhuh-mdb: Gotcha! I can bring that up in #ask-docs-platform. I’m personally just not super familiar with the different syntax we use to catch different versions and whatnot, but it’s probably better to have them take a look sooner rather than later 👍

ajhuh-mdb:
Hey @jeff-allen-mongo I know you already approved, but I wanted to go ahead and swap to the `${version}` placeholder sooner rather than later in case we run into more of these issues. Would you mind re-reviewing to make sure I got the syntax right?

jeff-allen-mongo:
I think this will work. Are you able to [test the redirects with mut](https://www.mongodb.com/docs/meta/redirects/#generating-redirects) so we can confirm? I believe the command to run is `mut-redirects config/redirects`

ajhuh-mdb:
That worked!

PR #618, opened 2024-04-10
DOCSP-38471 adding query on nested fields example

ltran-mdb2:
## DESCRIPTION

adds an example of querying on a nested field

## STAGING

https://preview-mongodbltranmdb2.gatsbyjs.io/compass/DOCSP-38471/query/filter/#match-by-nested-field

## JIRA

https://jira.mongodb.org/browse/DOCSP-38471

## BUILD LOG

https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=6616d9be3a5920a291ed9009

## Self-Review Checklist

- [x] Is this free of any warnings or errors in the RST?
- [x] Is this free of spelling errors?
- [x] Is this free of grammatical errors?
- [x] Is this free of staging / rendering issues?
- [x] Are all the links working?

### DIFF
--- a/source/query/filter.txt
+++ b/source/query/filter.txt
@@ -75,6 +75,9 @@ sample data into your MongoDB deployment, perform the following steps:
          {
             "name": "Andrea Le",
             "email": "andrea_le@fake-mail.com",
+            "school": {
+               "name": "Northwestern"
+            },
             "version": 5,
             "scores": [ 85, 95, 75 ],
             "dateCreated": { "$date": "2003-03-26" }
@@ -134,6 +137,9 @@ The query returns the following document:
       "_id": { "$oid": "5e349915cebae490877d561d" },
       "name": "Andrea Le",
       "email": "andrea_le@fake-mail.com",
+      "school": {
+         "name": "Northwestern"
+      },
       "version": 5,
       "scores": [ 85, 95, 75 ],
       "dateCreated": { "$date": "2003-03-26" }
@@ -183,6 +189,9 @@ The query returns the following documents:
          "_id": { "$oid": "5e349915cebae490877d561d" },
          "name": "Andrea Le",
          "email": "andrea_le@fake-mail.com",
+         "school": {
+            "name": "Northwestern"
+         },
          "version": 5,
          "scores": [ 85, 95, 75 ],
          "dateCreated": { "$date": "2003-03-26" }
@@ -295,6 +304,9 @@ The query returns the following documents:
          "_id": { "$oid": "5e349915cebae490877d561d" },
          "name": "Andrea Le",
          "email": "andrea_le@fake-mail.com",
+         "school": {
+            "name": "Northwestern"
+         },
          "version": 5,
          "scores": [ 85, 95, 75 ],
          "dateCreated": { "$date": "2003-03-26" }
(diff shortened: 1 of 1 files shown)


source/query/filter.txt
```diff
@@ -364,6 +364,35 @@ The query returns the following document:
...
+
+This example performs the following operations:
+
+- Adds a document with ``first`` and ``last`` subfields nested in the
+  ``name`` field
+- Finds the document with the ``last`` subfield of "Washington"
+
+Add the following document to your collection:
+
+.. code-block:: shell
```
jeff-allen-mongo, line 378: Can we use "javascript" highlighting here so we get the field name highlights?

source/query/filter.txt
```diff
@@ -364,6 +364,35 @@ The query returns the following document:
...
+
+   {
```
jeff-allen-mongo, line 380: Rather than adding a new document, could we maybe instead add an embedded field to the existing data at the top of the Examples section? The intro to the Examples section states that dataset can be used for the examples, so it might be confusing to have one example that uses a different sample doc.

We could keep it simple and do something like:

```
{
"name": "Andrea Le",
"email": "andrea_le@fake-mail.com",
"school": {
   "name": "NYU"
}
"version": 5,
"scores": [ 85, 95, 75 ],
"dateCreated": { "$date": "2003-03-26" }
}

File: source/query/filter.txt [hunk 3]
```diff
@@ -364,6 +364,35 @@ The query returns the following document:
       "dateCreated": { "$date": "2003-03-26" }
    }
 
+Match by Nested Field
```
jeff-allen-mongo, line 367: Generally we've been trying to use "embedded" as opposed to "nested" in our docs.

source/query/filter.txt
```diff
@@ -364,6 +364,35 @@ The query returns the following document:
...
+      "scores": [ 92, 87, 98 ],
+      "dateCreated": { "$date": "2003-11-22" }
+   }
+
+The following query filter uses the :query:`$eq` operator to find all
+documents with the ``name.last`` subfield of "Washington":
+
+.. code-block:: shell
+
+   { "name.last": { $eq: "Washington" } }
```
jeff-allen-mongo, line 393: I think we can remove $eq from this query. I think most users would write this query as { "name.last": "Washington" }

jeff-allen-mongo (commented):
Thanks for jumping on this so quickly! I left a couple comments and a question about whether we can integrate an embedded object into the existing sample docs.

ltran-mdb2:
Thanks for your feedback @jeff-allen-mongo ! This should be ready for another look

source/query/filter.txt
jeff-allen-mongo: Still have this question from last round of review.

source/query/filter.txt
```diff
@@ -359,11 +374,44 @@ The query returns the following document:
...
+.. code-block:: shell
+
+   { "school.name": "Northwestern" }
+
+
+The query returns the following document:
+
+.. code-block:: JSON
+   :copyable: false
+   :emphasize-lines: 4
```
jeff-allen-mongo, line 400: I think we want the emphasis on the school field? Currently it's on the email field.

jeff-allen-mongo (approved):
LGTM % a couple nits. Thanks @ltran-mdb2 !

source/query/filter.txt
jeff-allen-mongo: Actually ignore this! As discussed on Slack we can leave "shell" here to be consistent with the rest of hte page.

Anemy (approved):

PR #724, opened 2025-02-27
DOCSP-47211-default-sort-order-setting

ltran-mdb2:
## Self-Review Checklist

### DIFF
--- a/source/define-default-sort.txt
+++ b/source/define-default-sort.txt
@@ -1,0 +1,62 @@
+.. _compass-sort-query-bar:
+
+======================
+Set Default Sort Order
+======================
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+You can specify a default sort order documents that operations in the
+:ref:`query bar <compass-query-bar>` return. If you do not explicitly
+:ref:`specify a sort order in the query bar <compass-query-bar-sort>`, Compass
+applies the default sort order specified in the :guilabel:`Settings` panel. 
+
+About This Task
+---------------
+
+You can choose 1 of the following default sort options:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 20 80
+
+   * - Sort Key
+     - Description
+
+   * - $natural: 1 
+     - Sort in the natural order of documents
+
+   * - $natural: -1
+     - Sort in reverse natural order of documents
+
+   * - _id: 1
+     - Sort in ascending order by the ``_id`` field
+
+   * - _id: -1
+     - Sort in descending order by the ``_id`` field
+
+Procedure
+---------
+
+.. procedure:: 
+   :style: normal
+   
+   .. include:: includes/fact-navigate-to-settings.rst
+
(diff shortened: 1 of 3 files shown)


source/settings/set-default-sort.txt
```diff
@@ -0,0 +1,66 @@
+.. _compass-sort-query-bar:
+
+======================
+Set Default Sort Order
+======================
+
+.. default-domain:: mongodb
```
kennethdyer, line 7: I believe this directive is no longer functional.  It was being removed in some tickets earlier this week.

source/settings/set-default-sort.txt
```diff
@@ -0,0 +1,66 @@
...
+You can choose one of the following default sort options:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 20 80
+
+   * - Sort Key
+     - Description
+
+   * - $natural: 1
```
kennethdyer, line 35: Should these be literals?

source/settings/set-default-sort.txt
```diff
@@ -0,0 +1,66 @@
...
+   that you specify in the query overrides the default sort order.
+
+About This Task
+---------------
+
+You can choose one of the following default sort options:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 20 80
```
kennethdyer, line 30: Suggestion: We don't use it widely, but with these sort of lexical reference tables the `:stub-columns: 1` option sometimes helps with formatting for clarity.

kennethdyer (approved):
Thanks @ltran-mdb2, I left a few non-blocking suggestions for your consideration. LGTM!

source/settings/set-default-sort.txt
ltran-mdb2: Done

source/settings/set-default-sort.txt
ltran-mdb2: They're shown in normal text on the Compass UI  (shown in screenshot) but I will check with tech review
Screenshot 2025-03-03 at 12 33 57 PM

source/settings/set-default-sort.txt
ltran-mdb2: Done

djechlin-mongodb (approved):
I think it's good. Was wondering if you'd want to cover that these "defaults" aren't saved as part of the query when queries are saved, and furthermore that they're applied to already saved queries. I'm not sure if that level of detail is necessary.

source/settings/set-default-sort.txt
ltran-mdb2: @djechlin-mongodb what do you think here?

ltran-mdb2: Discussed with @djechlin-mongodb and we decided to leave these as non literals to match the Compass UI

ltran-mdb2:
> I think it's good. Was wondering if you'd want to cover that these "defaults" aren't saved as part of the query when queries are saved, and furthermore that they're applied to already saved queries. I'm not sure if that level of detail is necessary.

I added a line about how these defaults are not saved as part of saved queries!

